PR #7, opened 2025-08-21
feat: separate thread pools for monitor and work_impl to avoid task blocking.

Wang-1F:


yq33victor (approved):
LGTM

XuZhang99:
remove `WorkerImpl::get_active_activation_memory_async()` and use `WorkerImpl::get_active_activation_memory()` should also work.

Wang-1F:
> remove `WorkerImpl::get_active_activation_memory_async()` and use `WorkerImpl::get_active_activation_memory()` should also work.

We can retain the asynchronous semantics here for now, and later extend directly based on the wrapper if there are time-consuming operations.

PR #11, opened 2025-08-21
feat: modify the default values of devices and master_node_addr param…

JimHsiung:
…eters to simplify startup parameters.


yq33victor (approved):
LGTM

PR #17, opened 2025-08-22
refactor: refine active activation memory monitor.

XuZhang99:


xllm/core/runtime/vlm_engine.cpp
```diff
@@ -275,19 +275,10 @@ ForwardOutput VLMEngine::step(std::vector<Batch>& batches) {
 void VLMEngine::update_last_step_result(std::vector<Batch>& batch) {}
 
 std::vector<int64_t> VLMEngine::get_active_activation_memory() const {
```
yq33victor: In the engine, the interface needs to use asynchronous function, this ensures that requests sent to each worker are executed in parallel.

yq33victor (approved):
LGTM

PR #18, opened 2025-08-22
feat: add streaming tool call support for Qwen3.

yingxudeng:


yingxudeng:
This C++ implementation in the `xllm/function_call/partial_json_parser` directory was adapted from https://github.com/blaze2305/partial-json-parser (Go implementation) and https://github.com/promplate/partial-json-parser.git (Python implementation) with substantial modifications. I've also added comprehensive GTest cases to ensure its behavior remains consistent with the Python implementation. Should this be extracted as a third-party library, or would it be preferable to keep it within the xllm project?

yq33victor (approved):

PR #22, opened 2025-08-25
Feat: support kvcache store and host block manager.

Kang-Meng:


xllm/core/common/global_flags.cpp
```diff
@@ -181,6 +186,25 @@ DEFINE_bool(enable_atb_spec_kernel,
...
+DEFINE_bool(enable_kvcache_store, false, "whether to use kvcache store.");
+
+DEFINE_string(store_protocol, "tcp", "kvcache store protocol.");
+
+DEFINE_string(store_master_server_entry,
+              "tcp",
+              "address information of the store master service.");
+
+DEFINE_string(store_metadata_connstring,
+              "",
```
yq33victor, line 240: What is the difference between these two flags and `xservice_addr`/`etcd_addr`?

xllm/core/framework/block/block_manager_pool.cpp
```diff
@@ -144,8 +254,56 @@ void BlockManagerPool::allocate_shared(Sequence* sequence) {
...
+void BlockManagerPool::cache_host(Sequence* sequence) {
+  DCHECK(sequence != nullptr);
+
+  int32_t dp_rank = sequence->dp_rank();
+  size_t needed_block_num = (sequence->num_tokens() / options_.block_size() -
+                             sequence->host_kv_state().num_kv_blocks());
+
+  if (needed_block_num > 0) {
+    sequence->host_kv_state().add_kv_blocks(
+        host_block_managers_[dp_rank]->allocate(needed_block_num));
```
yq33victor, line 300: Don't need to handle the case when allocate failed ?

xllm/core/framework/kv_cache/kv_cache_store.cpp
```diff
@@ -0,0 +1,222 @@
...
+      value_tensor_one_layer.numel() * value_tensor_one_layer.element_size();
+
+  LOG(INFO) << "key_cache_size_per_layer: " << key_cache_size_per_layer_;
+  LOG(INFO) << "value_cache_size_per_layer: " << value_cache_size_per_layer_;
+
+  for (int layer = 0; layer < host_kv_caches_->size(); layer++) {
+    void* key_cache =
+        static_cast<char*>(host_kv_caches_->at(layer).get_k_cache().data_ptr());
+
+    auto register_k_result = client_ptr_->RegisterLocalMemory(
```
yq33victor: just only need to register memory when protocal is `rdma`

xllm/core/runtime/worker_impl.cpp
```diff
@@ -437,6 +546,44 @@ folly::SemiFuture<bool> WorkerImpl::pull_kv_blocks_async(
...
+  return kv_cache_store_->batch_put(cache_block_info);
+}
+
+folly::SemiFuture<uint32_t> WorkerImpl::offload_kv_blocks_to_store_async(
+    const std::vector<CacheBlockInfo>& cache_block_info) {
+  folly::Promise<uint32_t> promise;
+  auto future = promise.getSemiFuture();
+  threadpool_.schedule(
+      [this, &cache_block_info, promise = std::move(promise)]() mutable {
+        if (this->kv_cache_store_ == nullptr) {
```
yq33victor, line 657: Even though a thread pool is used here, it is still a synchronous interface. This is because the thread count (thread num) of the thread pool in workimpl is set to 1, so subsequent step_async operations will still be blocked in the thread pool.

xllm/core/distributed_runtime/remote_worker.cpp
```diff
@@ -429,6 +444,31 @@ folly::SemiFuture<bool> RemoteWorker::pull_kv_blocks_async(
   return future;
 }
 
+folly::SemiFuture<uint32_t> RemoteWorker::load_kv_blocks_from_store_async(
+    const std::vector<CacheBlockInfo>& cache_block_info) {
+  folly::Promise<uint32_t> promise;
+  auto future = promise.getSemiFuture();
+  threadpool_.schedule(
+      [this, &cache_block_info, promise = std::move(promise)]() mutable {
```
yq33victor: There's no need to use an asynchronous interface here, using one would actually cause synchronous execution with the `Step`. Here's how the call chain works:
In the ContinuousScheduler::add_request function, prepare_cache_async is called, which then invokes the engine_->load_kv_blocks_from_store_async function. The engine has added a new thread pool (threadpool_) for asynchronous scheduling, and everything works fine up to this point.
However, when the LLMEngine::load_kv_blocks_from_store_async function calls the client's load_kv_blocks_from_store_async, it will invoke the RemoteWorker::load_kv_blocks_from_store_async function of the remote worker. The threadpool_ in the remote worker has only 1 thread, which creates an implicit synchronization issue with the main thread's `Step`.
Therefore, it's recommended that the engine rename threadpool_ to async_load_threadpool (or a similar name) specifically for use by prepare_cache_async, and synchronous interfaces should be used in all subsequent parts of the call chain.

xllm/core/runtime/worker_impl.cpp
```diff
@@ -437,6 +546,44 @@ folly::SemiFuture<bool> WorkerImpl::pull_kv_blocks_async(
                                                   dst_blocks);
 }
 
+folly::SemiFuture<uint32_t> WorkerImpl::load_kv_blocks_from_store_async(
+    const std::vector<CacheBlockInfo>& cache_block_info) {
+  folly::Promise<uint32_t> promise;
+  auto future = promise.getSemiFuture();
+  threadpool_.schedule(
+      [this, &cache_block_info, promise = std::move(promise)]() mutable {
```
yq33victor, line 633: Similarly, the number of threads in the thread pool here is 1, so an additional thread needs to be allocated for this purpose. Create a separate thread pool.

xllm/core/distributed_runtime/worker_service.cpp
```diff
@@ -215,6 +216,23 @@ void WorkerService::PullKVCache(::google::protobuf::RpcController* controller,
   return;
 }
 
+void WorkerService::LoadKVCacheFromStore(
+    ::google::protobuf::RpcController* controller,
+    const ::xllm::proto::CacheBlockInfos* req,
+    ::xllm::proto::StoreResponse* resp,
+    ::google::protobuf::Closure* done) {
+  threadpool_.schedule([this, controller, req, resp, done]() mutable {
+    brpc::ClosureGuard done_guard(done);
```
yq33victor: worker service has a `ThreadPool threadpool_{5};`, so it is ok here

xllm/core/runtime/worker.cpp
```diff
@@ -139,6 +139,11 @@ folly::SemiFuture<bool> Worker::pull_kv_blocks_async(
                                      dst_blocks);
 }
 
+folly::SemiFuture<uint32_t> Worker::load_kv_blocks_from_store_async(
+    const std::vector<CacheBlockInfo>& cache_block_info) {
+  return impl_->load_kv_blocks_from_store_async(cache_block_info);
```
yq33victor, line 159: In worker_impl:
Similarly, the number of threads in the worker_impl thread pool here is 1, so an additional thread needs to be allocated for this purpose. Create a separate thread pool.

xllm/core/runtime/worker_impl.cpp
```diff
@@ -90,6 +91,48 @@ bool WorkerImpl::allocate_kv_cache(
...
 
+  allocate_host_kv_cache(kv_cache_shape);
+  status_ = Status::READY;
+  return true;
+}
+
+bool WorkerImpl::allocate_host_kv_cache(
+    const std::vector<std::vector<int64_t>>& device_kv_cache_shape) {
+  if (options_.host_blocks_factor() <= 0.0) {
+    return true;
```
yq33victor, line 134: nit: When comparing floats, there may be some precision issues. Therefore, to be safe, you can choose a magic number such as 0.00001.
`if (options_.host_blocks_factor() <= 0.00001)`

xllm/core/framework/block/block_manager_pool.cpp
```diff
@@ -115,6 +173,25 @@ bool BlockManagerPool::allocate(Sequence* sequence, size_t num_tokens) {
...
+  size_t host_cache_token_num = sequence->host_kv_state().kv_cache_tokens_num();
+  if (hbm_cache_token_num < host_cache_token_num) {
+    auto hbm_blocks = sequence->kv_state().kv_blocks();
+    auto host_blocks = sequence->host_kv_state().kv_blocks();
+
+    for (int i = hbm_cache_token_num / options_.block_size();
+         i < host_cache_token_num / options_.block_size();
+         i++) {
+      copy_in_cache_contents_[sequence->dp_rank()].emplace_back(
+          hbm_blocks[i].id(),
```
yq33victor, line 202: when to allocate hbm_blocks for `copy in block`

xllm/core/framework/block/block_manager_pool.cpp
```diff
@@ -115,6 +173,25 @@ bool BlockManagerPool::allocate(Sequence* sequence, size_t num_tokens) {
...
+          host_blocks[i].id(),
+          host_blocks[i].get_immutable_hash_value());
+    }
+    sequence->kv_state().incr_kv_cache_tokens_num(host_cache_token_num -
+                                                  hbm_cache_token_num);
```
yq33victor, line 207: Here, what we've allocated is host KV cache, and no HBM blocks have been allocated. Therefore, we cannot directly call the function sequence->kv_state().incr_kv_cache_tokens_num(...).

xllm/core/scheduler/continuous_scheduler.cpp
```diff
@@ -102,7 +104,9 @@ void ContinuousScheduler::handle_prefill_requests(
         continue;
       }
 
-      size_t num_tokens = prefill_sequence->num_tokens();
+      prefill_sequence->sync_result();
```
yq33victor, line 179: When we call the sync_result() function, if the future is not yet ready, it won't wait. Is this a reasonable logic?

xllm/core/common/global_flags.cpp
Kang-Meng: xservice_addr: Specifies the address of a single xllm service instance (typically used in legacy deployments)
etcd_addr: Specifies the address of the etcd service discovery cluster used to locate the active master in multi-service deployments

Kang-Meng: - xservice_addr: Specifies the address of a single xllm service instance (typically used in legacy deployments).

- etcd_addr: Specifies the address of the etcd service discovery cluster used to locate the active master in multi-service deployments.

The system uses etcd for service discovery and leader election. Xllm should connect to etcd (via etcd_addr) to dynamically discover the current master xllm-service.

xllm/core/framework/block/block_manager_pool.cpp
Kang-Meng: In the BlockManagerPool::allocate function, it first allocates HBM blocks. Then, it checks whether the number of tokens cached in the host exceeds those in HBM cache. Only when this condition is met will incr_kv_cache_tokens_num be executed. For the excess tokens, a synchronous HtoD copy will be performed before executing the upcoming step to ensure the tensors corresponding to blocks in HBM cache are correct during actual computation.

xllm/core/framework/block/block_manager_pool.cpp
Kang-Meng: refer to block_manager_pool.cpp:152 block_manager_pool.cpp:168

yq33victor (dismissed):

PR #25, opened 2025-08-25
feat: add compilation support for mlu.

XuZhang99:


xllm/core/common/device_memory.cpp
```diff
@@ -3,6 +3,11 @@
 #include <glog/logging.h>
```
yq33victor, line 3: If there are too many diff in a file, we can write an additional new file for mlu, that we can manager npu/mlu in the Makefile. This will make the overall structure looks cleaner.

XuZhang99: If we write a new device_memory for mlu, we need macro to determine which device_memory to use in `WorkerImpl::estimate_kv_cache_capacity()`.

yq33victor: > If we write a new device_memory for mlu, we need macro to determine which device_memory to use in `WorkerImpl::estimate_kv_cache_capacity()`.

we determine it in makefile. Just split them into different files, but they have the same class name, function etc.

The premise is when most code in a file cannot be reused.

yq33victor (approved):

liutongxuan (approved):

PR #27, opened 2025-08-25
feat: refactor layer module to support multiple platforms.

liujinguang0125:


xllm/core/layers/base_layer.h
```diff
@@ -0,0 +1,129 @@
...
+#include <string>
+#include <vector>
+
+#include "framework/context.h"
+#include "framework/kv_cache/kv_cache.h"
+#include "framework/model/model_input_params.h"
+#include "framework/state_dict/state_dict.h"
+
+namespace xllm {
+namespace layer {
```
liutongxuan, line 36: 建议这里可能不用加layer这个namespace了，感觉有点多余了

xllm/core/layers/qwen2_decoder_layer.h
```diff
@@ -0,0 +1,24 @@
+#pragma once
+
+// #ifdef WITH_ASCEND
```
yq33victor: nit: delete these comments about `WITH_ASCEND` :(

xllm/core/layers/CMakeLists.txt
```diff
@@ -19,14 +19,47 @@ cc_library(
     torch
 )
 
+cc_library(
```
a120092009, line 27: 建议暂缓合入

xllm/core/layers/qwen2_decoder_layer.h
XuZhang99: already removed

xllm/core/layers/base_layer.h
XuZhang99: already removed namespace layer

XuZhang99: The namespace layer cannot be deleted yet to prevent naming conflicts among some LayerImpl classes in the model class. Subsequent refactoring will remove the LayerImpl  classes from the model and eliminate the namespace layer altogether.

xllm/core/layers/npu/npu_base_layer.cpp
```diff
@@ -0,0 +1,88 @@
...
+    LOG(ERROR) << " setup layer node fail, not call execute";
+    return st;
+  }
+
+  if (node.workspaceSize > 0) {
+    node.workspace = workspace.get_workspace_buffer(node.workspaceSize);
+  }
+
+  run_task_func_(name_ + std::to_string(node_id), [=]() {
+    return execute_plan(
```
yq33victor, line 50: it's better to modify `std::function<void(const std::string&, std::function<int()>)> run_task_func_;`
to 
`std::function<atb::Status(const std::string&, std::function<int()>)> run_task_func_;` ?

that we can return 
```
return run_task_func_(name_ + std::to_string(node_id), [=]() {....});
```

xllm/core/layers/npu/npu_deepseek_v2_decoder_layer_impl.cpp
```diff
@@ -679,16 +683,17 @@ int DeepseekV2DecoderImpl::get_mapped_index(
   return it->second;
 }
 
-std::string DeepseekV2DecoderImpl::get_expert_shm_key(int32_t layer_id,
-                                                      int32_t expert_index,
-                                                      std::string suffix) {
+std::string NpuDeepseekV2DecoderLayerImpl::get_expert_shm_key(
+    int32_t layer_id,
+    int32_t expert_index,
+    std::string suffix) {
```
yq33victor: nit: 
`std::string suffix` -> `const std::string& suffix`

RobbieLeung:
@XuZhang99 How about we split the NPU layers into two groups to simplify maintenance?
- Basic op - like Embedding, RMS Norm.
- Fused subgraphs for decode layers (Qwen, Deepseek etc.)

Directory would look like:

```
layers/
├── base/
│   ├── npu_word_embedding_impl.h
│   └── npu_rms_norm_impl.h
│
└── fusion/
    ├── npu_qwen3_decoder_layer_impl.h
    └── npu_deepseek_v2_decoder_layer_impl.h
```

yq33victor (approved):
LGTM

PR #29, opened 2025-08-26
refactor: unify the schedulers for prefill and decode into one.

JimHsiung:


xllm/core/scheduler/continuous_scheduler.cpp
```diff
@@ -407,6 +407,9 @@ std::vector<Batch> ContinuousScheduler::prepare_batch() {
   std::vector<std::shared_ptr<Request>> finished_requests;
   for (auto it = running_requests_.rbegin(); it != running_requests_.rend();
        ++it) {
+    if (*it == nullptr) {
+      continue;
```
yq33victor, line 411: emmm.. in which case it is a nullptr ?

JimHsiung: Currently, DisaggPDScheduler directly reuses the prepare_batch from ContinuousScheduler. In prefill_send_first_generation, it checks whether running_requests_ need to be sent to decode, and if so, it sets the corresponding request in running_requests_ to nullptr. In the previous logic, DisaggPDScheduler would send all requests to decode in prefill_send_first_generation, then in the overridden prepare_batch, it would directly clear all elements in running_requests_ at the beginning of each step.

PR #45, opened 2025-08-27
bugfix: fix compile issuse for mlu.

XuZhang99:


zhang-minchao (dismissed):

setup.py
```diff
@@ -189,7 +189,10 @@ def set_npu_envs():
 
 # TODO(mlu): set mlu environment variables
 def set_mlu_envs():
-    pass
+    os.environ["PYTHON_INCLUDE_PATH"] = get_python_include_path()
+    os.environ["PYTHON_LIB_PATH"] =  get_torch_root_path()
+    os.environ["LIBTORCH_ROOT"] = get_torch_root_path()
+    os.environ["LIBTORCH_ROOT"] = get_torch_root_path()
```
linkerzhang: the two lines set the same one?

xllm/core/framework/parallel_state.cpp
```diff
@@ -172,7 +180,15 @@ std::vector<std::unique_ptr<ProcessGroup>> ProcessGroup::create_process_groups(
   }
   return process_groups;
 }
+#elif defined(USE_MLU)
+// TODO(mlu): implement create_process_groups for mlu
+std::vector<std::unique_ptr<ProcessGroup>> ProcessGroup::create_process_groups(
+    const std::vector<torch::Device>& devices) {
+  return {};
```
linkerzhang, line 202: not good to check-in an empty function, remove it or finish it please.

XuZhang99: this function needs to be implemented by the MLU team, so I leave it as a TODO.

setup.py
XuZhang99: These four lines of code were copied from set_npu_envs(), and the duplication issue existed before—I forgot to check it. The last line is removed now.

yq33victor (approved):

PR #48, opened 2025-08-28
feat: support multi-priority and on/offline unified request schedule.

weizhehuang0827:
implement multi-priority and on/offline unified request schedule.


weizhehuang0827:
Note that a simpler alternative strategy is to continue using `deque` to implement decode queue and sort it directly each time when scheduling, which seems to have higher complexity in theory.

xllm/core/framework/request/request.cpp
```diff
@@ -32,13 +32,19 @@ Request::Request(const std::string& request_id,
                  const std::string& x_request_id,
                  const std::string& x_request_time,
                  const RequestState& state,
-                 const std::string& service_request_id)
+                 const std::string& service_request_id,
+                 bool offline,
+                 int32_t slo_ms,
+                 xllm::proto::Priority priority)
```
yq33victor: Create a new `enum Priority` to replace `xllm::proto::Priority`, we'd better avoid using `protocol buffer types` during serving.

xllm/core/framework/request/request_params.h
```diff
@@ -124,6 +125,12 @@ struct RequestParams {
   std::vector<xllm::JsonTool> tools;
   std::string tool_choice = "auto";
   bool has_tools() const { return !tools.empty(); }
+
+  bool offline = false;
+
+  int32_t slo_ms = 0;
+
+  xllm::proto::Priority priority = xllm::proto::Priority::NORMAL;
```
yq33victor: As mentioned above, a new `enum Priority` type should be used here.

xllm/core/scheduler/disagg_pd_scheduler.cpp
```diff
@@ -179,7 +184,19 @@ bool DisaggPDScheduler::add_request(std::shared_ptr<Request>& request) {
 void DisaggPDScheduler::dispatch_requests() {
   while (true) {
     std::vector<std::shared_ptr<Request>> requests;
-    std::shared_ptr<Request> request = prefill_request_queue_.pop();
+
+    auto poped_result = prefill_request_queue_.try_pop();
+    // OPTIMIZE 之后改为：多次尝试读取在线 prefill
```
yq33victor: nit: emmm... comments in English.  :)

xllm/core/common/global_flags.cpp
```diff
@@ -199,3 +199,9 @@ DEFINE_string(etcd_addr, "", "etcd adderss for save instance meta info");
 DEFINE_bool(enable_service_routing, false, "whether to use etcd.");
 
 DEFINE_int32(heart_beat_interval, 3, "heart beat interval");
+
+DEFINE_string(priority_strategy, "FCFS", "priority strategy for requests");
+
+DEFINE_bool(enable_on_preempt_off,
```
yq33victor: nit:  modify the flag name, as it may be mistaken for `enable`, `on_preempt`,  `off` at first glance.

use `enable_online_preempt_offline` or some names else.

xllm/core/framework/block/block_manager_impl.cpp
```diff
@@ -73,6 +74,33 @@ void BlockManagerImpl::deallocate(const Slice<Block>& blocks) {
...
+       it != running_queue_to_evict->rend();
+       ++it) {
+    std::shared_ptr<Request> request_to_preempt = *it;
+    num_request_to_evict++;
+    // count the number of blocks belong to the request
+    for (const auto& seq : request_to_preempt->sequences()) {
+      num_blocks_can_evict += seq->kv_state().num_kv_blocks();
+    }
+    if ((num_blocks_needed <= num_blocks_can_evict) ||
+        has_enough_blocks(num_blocks_needed - num_blocks_can_evict)) {
```
yq33victor: We don't need to call `has_enough_blocks` every time. This is because before calling `check_if_enough_to_evict`, the allocate operation must have failed. Additionally, the entire execution code runs serially with no concurrency (in the PD separation scenario, there is concurrency when allocating blocks for remote prompts, but this only involves block allocation and no block release). Therefore, the number of free blocks will not increase. 

The `has_enough_blocks` function call `prefix_cache` eviction logic every time, and this may cost to much.

Based on the above considerations, we can place the function `check_if_enough_to_evict` directly in the contiguous scheduler file. This is because we only need to judge whether the total `num_kv_blocks` in `running_queue_to_evict` meets the size requirement of `num_request_to_evict`. 
BlockManagerImpl is inherently used for managing block allocations, it is not very appropriate to pass parameters such as `DecodePriorityQueue` to it.

xllm/core/scheduler/continuous_scheduler.cpp
```diff
@@ -182,20 +245,14 @@ void ContinuousScheduler::handle_prefill_requests(
...
-    return;
-  }
-
-  // Handle decoding requests.
-  // no prefill request, schedule the decode requests in the running priority
-  // queue
-  while (!running_queue_.empty() &&
+    size_t& num_offd_preempt_off_requests,
+    size_t& num_ond_preempt_on_requests,
+    size_t& num_ond_preempt_off_requests,
```
yq33victor: nit: `offd` and `ond` , It’s not immediately known what it means.  :(

xllm/core/scheduler/continuous_scheduler.cpp
```diff
@@ -182,20 +245,14 @@ void ContinuousScheduler::handle_prefill_requests(
 void ContinuousScheduler::handle_decode_requests(
     size_t& remaining_token_budget,
     size_t& remaining_seq_budget,
-    size_t& num_preempted_requests) {
-  // Do nothing: have new prefill requests to handle, or have no running
-  // requests
-  if (!running_sequences_.empty() || running_queue_.empty()) {
```
yq33victor, line 189: emmm... What is the purpose of removing this logic? In `ContinuousScheduler`, prefill will be will be executed with high priority.

xllm/core/scheduler/continuous_scheduler.cpp
```diff
@@ -220,9 +277,9 @@ void ContinuousScheduler::handle_decode_requests(
         has_enough_budget = false;
         break;
       }
-
+      // sequence token already appended
       size_t updated_num_tokens =
-          sequence->num_tokens() + options_.num_speculative_tokens() + 1;
+          sequence->num_tokens() + options_.num_speculative_tokens();
```
yq33victor, line 308: the target new generated tokens is `options_.num_speculative_tokens() + 1`

xllm/core/scheduler/continuous_scheduler.cpp
```diff
@@ -537,7 +691,8 @@ std::vector<Batch> ContinuousScheduler::schedule_request(
       return batch;
     }
 
-    if (!waiting_priority_queue_.empty() || !running_queue_.empty()) {
+    if (!waiting_priority_queue_.empty() || !running_queue_->empty() ||
+        !waiting_priority_queue_offline_.empty()) {
```
yq33victor: Add running_queue_offline_ ?

xllm/core/scheduler/disagg_pd_scheduler.cpp
```diff
@@ -179,7 +184,19 @@ bool DisaggPDScheduler::add_request(std::shared_ptr<Request>& request) {
...
+    // 请求，只有较长时间未获取在线请求时，才读取离线 prefill 请求。
+    if (!poped_result.has_value()) {
+      poped_result = prefill_request_queue_offline_.try_pop();
+      if (!poped_result.has_value()) {
+        // no offline request, sleep for a while and try again
+        absl::SleepFor(absl::Milliseconds(100));
```
yq33victor: this maybe block online request.

A more appropriate strategy is to design the pop interface as a blocking interface with timeout waiting (for example, it automatically returns null after 100ms), and then use try_pop to read offline requests. This way, what we block are offline requests rather than online requests.

```
std::shared_ptr<Request> poped_result = prefill_request_queue_.pop();
if (poped_result.has_value()) {
   // nothing
} else {
   poped_result = prefill_request_queue_offline_.try_pop();
}
```

pop function:
```
    absl::optional<T> pop(absl::Duration timeout) {
        absl::MutexLock lock(&mutex_);

        bool has_value = mutex_.AwaitWithTimeout(
            absl::Condition(
                +[](std::queue<T>* queue) { return !queue->empty(); },
                &queue_
            ),
            timeout
        );

        if (!has_value) {
            return absl::nullopt;
        }

        T value = std::move(queue_.front());
        queue_.pop();
        return value;
    }
```

xllm/core/scheduler/continuous_scheduler.cpp
weizhehuang0827: In my test script, I confirm the new token has been added to sequence `num_tokens`. There is no need to `+1` again. There is a logic difference between `ContinuousScheduler` and `ChunkedPrefillScheduler` here before.

xllm/core/scheduler/disagg_pd_scheduler.cpp
weizhehuang0827: Sorry :(

xllm/core/scheduler/continuous_scheduler.cpp
weizhehuang0827: just want to make name short and align with `num_preempt_requests` orz. OK I will change it to whole name.

xllm/core/scheduler/continuous_scheduler.cpp
weizhehuang0827: Yes!

xllm/core/scheduler/continuous_scheduler.cpp
weizhehuang0827: I move this logic to `prepare_batch` and remove the redundant second `empty`. I hope to wrap the two `handle_decode_requests` to ensure that only prefill or decode is processed.

PR #55, opened 2025-08-29
bugfix: fix prefix cache evict bug.

yq33victor:


xllm/core/scheduler/chunked_prefill_scheduler.cpp
```diff
@@ -166,7 +166,7 @@ void ChunkedPrefillScheduler::handle_running_queue_requests(
       }
 
       // if (sequence->if_cache_block_for_prefill()) {
-      //   block_manager_->cache(sequence.get());
+      //   block_manager_pool_->cache(sequence.get());
```
linkerzhang, line 169: remove these useless codes please.

xllm/core/scheduler/zero_eviction_scheduler.h
```diff
@@ -100,7 +100,7 @@ class BlockCapacityGuard {
   std::vector<SequenceStatus> get_running_sequence_status();
 
  private:
-  BlockManagerPool* block_manager_;
+  BlockManagerPool* block_manager_pool_;
```
linkerzhang, line 103: it would be better to use smart pointer.

xllm/core/framework/prefix_cache/prefix_cache_hash_murmur3.cpp
```diff
@@ -233,6 +233,7 @@ size_t PrefixCacheHashMurmur3::evict(size_t n_blocks) {
     delete del_node;
     ++evict_count;
     --num_blocks_;
+    ++i;
```
linkerzhang, line 236: why move the."++i" here?

xllm/core/framework/prefix_cache/prefix_cache_hash_sha256.cpp
```diff
@@ -185,6 +185,7 @@ size_t PrefixCacheHashSha256::evict(size_t n_blocks) {
     delete del_node;
 
     --num_blocks_;
+    ++evict_count;
```
linkerzhang, line 188: same as above.

xllm/core/scheduler/chunked_prefill_scheduler.cpp
```diff
@@ -77,7 +77,7 @@ void ChunkedPrefillScheduler::handle_abnormal_request(
           << "Running queue size is not 1, there maybe a bug of request "
              "preemption logic. running_queue_.size ="
           << running_queue_.size();
-      if (util::sum(block_manager_->num_used_blocks()) !=
+      if (util::sum(block_manager_pool_->num_used_blocks()) !=
```
linkerzhang, line 80: can block_manager_pool_ be nullptr?

xllm/core/scheduler/continuous_scheduler.h
```diff
@@ -126,7 +126,7 @@ class ContinuousScheduler : public Scheduler {
   Engine* engine_;
 
   // the block manager to manage the cache blocks
-  BlockManagerPool* block_manager_;
+  BlockManagerPool* block_manager_pool_;
```
linkerzhang, line 129: it would be better to move "renaming" to be a separate PR, to make the bugfix PR clear.

xllm/core/framework/prefix_cache/prefix_cache_hash_murmur3.cpp
yq33victor: Using ++i in the for loop will cause a code logic error, it leads to an incorrect counting.

xllm/core/scheduler/chunked_prefill_scheduler.cpp
yq33victor: `block_manager_pool_` will  never be a nullptr.

xllm/core/framework/prefix_cache/prefix_cache_hash_murmur3.cpp
weizhehuang0827: Because we want to evict a total of `n_blocks` blocks here, but we will **skip** the blocks that are being shared. Moving `++i` here prevents the `for` loop from stopping before evicting `n_blocks` blocks. 
Maybe `i` is redundant here. we can directly use `while(evict_count<n_blocks){...;++evict_count}`.

xllm/core/scheduler/continuous_scheduler.h
yq33victor: > it would be better to move "renaming" to be a separate PR, to make the bugfix PR clear.

ok, I will split the pr into two.

JimHsiung (approved):

PR #62, opened 2025-09-01
docs: update README model support.

wly-115:
docs: update README model support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,7 @@ Supported models list:
 - Kimi-k2
 - Llama2/3
 - MiniCPM-V
+- MiMo-VL
 - Qwen2/2.5/QwQ
 - Qwen2.5-VL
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -99,6 +99,7 @@ Supported models list:
 - Kimi-k2
 - Llama2/3
 - MiniCPM-V
+- MiMo-VL
```
yq33victor, line 102: the model file is not commited ?

wly-115: @yq33victor model type is same with qwen2.5 vl，just update readme

liutongxuan, line 102: Should MiMo be added as well？

wly-115: @liutongxuan MiMo=Qwen2+MTP, model_type=MiMo，it needs a model file.

liutongxuan (approved):

PR #63, opened 2025-09-01
bugfix: ensure xllm ops in the third party directory rebuilds when updated.

LMX-xin:


CMakeLists.txt
```diff
@@ -24,10 +24,19 @@ if(USE_NPU)
 
   option(PRECOMPILE_XLLM_OPS "Precompile xllm ops before main build" ON)
```
liutongxuan: it seems we don't need the PRECOMPILE_XLLM_OPS. just remove it.

LMX-xin: okk

LMX-xin: > it seems we don't need the PRECOMPILE_XLLM_OPS. just remove it.

PRECOMPILE_XLLM_OPS has been deleted ,and the USE_A2 macro has been used instead.

yq33victor (approved):

PR #72, opened 2025-09-02
feat: support multi-priority and on/offline for ChunkedPrefill.

weizhehuang0827:
- support multi-priority and on/offline for ChunkedPrefill.
- fix expand_sequence.
- refactor some in scheduler.


yq33victor (approved):
LGTM

xllm/core/scheduler/chunked_prefill_scheduler_test.cpp
```diff
@@ -129,6 +131,51 @@ std::vector<std::shared_ptr<Request>> generate_request(
   return requests;
 }
 
+std::vector<std::shared_ptr<Request>> generate_priority_request(
+    const std::vector<int32_t>& prompt_lens,
+    const std::vector<int32_t>& max_tokens,
+    const std::vector<int32_t>& offlines,
```
XuZhang99: it should be `const std::vector<bool>& offlines`

liutongxuan (approved):

PR #92, opened 2025-09-04
feat: support continuous kvcache using virtual memory.

XuZhang99:


xllm/core/scheduler/kv_cache_manager_client.h
```diff
@@ -0,0 +1,62 @@
...
+  std::vector<size_t> num_blocks_in_prefix_cache() const;
+  std::vector<size_t> num_free_blocks() const;
+  std::vector<size_t> num_used_blocks() const;
+  double kv_cache_utilization() const;
+
+  const BlockManagerPool::Options& options() const;
+
+ private:
+  // these two pointers must be one null and one non-null
+  BlockManagerPool* block_manager_pool_;
```
liutongxuan: Can you build a ContinousBlockManagerPool class inherits from BlockManagerPool? In this way, this Client is not needed

xllm/core/distributed_runtime/remote_worker.cpp
```diff
@@ -376,6 +406,42 @@ folly::SemiFuture<bool> RemoteWorker::allocate_kv_cache_async(
...
+        options[0].head_size());
+    xtensor_options_vec.mutable_key_options()->set_max_context_len(
+        options[0].max_context_len());
+    xtensor_options_vec.mutable_key_options()->set_max_seqs_per_batch(
+        options[0].max_seqs_per_batch());
+    xtensor_options_vec.mutable_value_options()->set_num_kv_heads(
+        options[1].num_kv_heads());
+    xtensor_options_vec.mutable_value_options()->set_head_size(
+        options[1].head_size());
+    xtensor_options_vec.mutable_value_options()->set_max_context_len(
```
yq33victor, line 435: call function `allocate_continuous_kv_cache` directly?

xllm/core/framework/batch/batch_input_builder.h
```diff
@@ -67,6 +67,10 @@ class BatchInputBuilder {
                            uint32_t n_kv_cache_tokens,
                            uint32_t seq_len,
                            uint32_t q_seq_len);
+  void setup_continuous_kv_cache_info(Sequence* sequence,
```
yq33victor, line 71: function `setup_continuous_kv_cache_info` is not used ?

xllm/core/framework/batch/batch_input_builder.cpp
```diff
@@ -346,6 +369,13 @@ ForwardInput BatchInputBuilder::state_to_forward_input() {
     input_params.input_embedding = torch::cat(input_embeddings_vec_);
   }
 
+  if (FLAGS_enable_continuous_kvcache) {
+    input_params.new_cache_slots =
+        torch::tensor(state_.new_cache_slot_offsets, torch::kInt64);
```
yq33victor, line 390: type is `kInt` above in line 355 ?

xllm/core/framework/kv_cache/kv_cache.h
```diff
@@ -19,24 +19,38 @@ limitations under the License.
...
-    return !key_cache_.defined() || !value_cache_.defined();
+    return FLAGS_enable_continuous_kvcache
+               ? (key_xtensor_ == nullptr || value_xtensor_ == nullptr)
+               : (!key_cache_.defined() || !value_cache_.defined());
   }
 
  private:
   torch::Tensor key_cache_;
   torch::Tensor value_cache_;
+
```
yq33victor, line 53: nit: maybe it's better create a new class like `class ContinuousKVcache`.

Anyway, it's ok currently.  :)  ignore.

xllm/core/framework/xtensor/xtensor_manager.cpp
```diff
@@ -0,0 +1,246 @@
...
+      std::make_pair(std::move(multi_layer_kv_xtensor.first),
+                     std::move(multi_layer_kv_xtensor.second));
+}
+
+int32_t PageManager::allocate_seq_id() {
+  int32_t seq_id = -1;
+  // TODO: refine this
+  seq_id = multi_layer_kv_xtensor_.first->allocate_seq_id();
+  seq_id = multi_layer_kv_xtensor_.second->allocate_seq_id();
+}
```
yq33victor, line 44: no returned value ?

xllm/core/framework/page/page_manager_client.cpp
```diff
@@ -0,0 +1,64 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#include "page_manager_client.h"
+
+namespace xllm {
+
+bool PageManagerClient::allocate(int32_t& seq_id, size_t num_tokens) {
```
yq33victor: `PageManagerClient` is just the wrapper of `page_manager_`, could we use `page_manager_` directly ?

xllm/core/runtime/engine.h
```diff
@@ -119,6 +125,9 @@ class Engine {
   // block manager
   std::unique_ptr<BlockManagerPool> block_manager_pool_;
 
+  // page manager
+  std::unique_ptr<PageManagerPool> page_manager_pool_;
```
yq33victor: can we create a base `ManagerPool` class, `BlockManagerPool` and `PageManagerPool` can inherit from the base class. Then here we can only define a member `std::unique_ptr<ManagerPool> manager_pool_;`

xllm/core/framework/page/page_manager_server.cpp
```diff
@@ -0,0 +1,151 @@
...
+#include <brpc/channel.h>
+
+#include "common/global_flags.h"
+#include "page_manager.h"
+#include "page_manager_service.h"
+#include "server/xllm_server_registry.h"
+#include "util/net.h"
+
+namespace xllm {
+void PageManagerServer::create_server(const page::Options& options,
```
yq33victor: nit: we'd better put these files here to their respective directories according to their functions.
like: `page_manager_pool.h` to `framework/block/`, `page_manager_server.h` to `core/distribute_runtime`, `service` ... etc. we can add a subdirectory under these directories like `page/`.

Anyway, this can be refactored later, currently is ok.

xllm/core/framework/request/sequence.h
```diff
@@ -247,6 +247,12 @@ class Sequence final {
 
   void reset();
 
+  // set sequence id
+  void set_seq_id(int32_t seq_id) { seq_id_ = seq_id; }
```
yq33victor, line 260: nit: can we use `index_` in `class Sequence` directly?

xllm/core/scheduler/kv_cache_manager_client.cpp
```diff
@@ -0,0 +1,161 @@
...
+==============================================================================*/
+
+#include "kv_cache_manager_client.h"
+
+#include <glog/logging.h>
+
+namespace xllm {
+
+KVCacheManagerClient::KVCacheManagerClient(BlockManagerPool* block_manager_pool)
+    : block_manager_pool_(block_manager_pool) {}
```
yq33victor: Here, after we create a base class `ManagerPool`, there will be no need to determine whether it is a `block` or a `page`. We can directly handle it with the `base class pointer`. And the `KVCacheManagerClient` will also become unnecessary.

xllm/core/framework/request/sequence.h
XuZhang99: the meaning of `index_` and `seq_id_` is different. `index_` is the index of the sequence in the **request**, but `seq_id_` the index of the sequence in the **batch**.

xllm/core/framework/batch/batch_input_builder.h
XuZhang99: it should be used in func `process_single_sequence`, but i forgot...

xllm/core/framework/page/page_manager_client.cpp
XuZhang99: it's like `WorkerImpl` and `WorkerClient`, i think `WorkerClient` is used for single node serving, so i use `PageManagerClient` for single node serving.

xllm/core/framework/batch/batch_input_builder.cpp
XuZhang99: `new_token_slot_id` is `int32_t`, but `new_cache_slot_offset` is `int64_t`

yq33victor (approved):

PR #93, opened 2025-09-04
feat: support image url input for VLM.

wly-115:
feat: support image url input for VLM.

### DIFF
--- a/xllm/core/framework/request/mm_input_helper.cpp
+++ b/xllm/core/framework/request/mm_input_helper.cpp
@@ -15,6 +15,11 @@ limitations under the License.
 
 #include "mm_input_helper.h"
 
+#include <brpc/channel.h>
+#include <brpc/controller.h>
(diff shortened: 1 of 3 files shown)


xllm/core/framework/request/mm_input_helper.cpp
```diff
@@ -41,6 +45,73 @@ class OpenCVImageDecoder {
...
+    const butil::IOBuf& io = cntl_->response_attachment();
+    data = io.to_string();
+    return true;
+  }
+
+  bool fetch_data(const std::string& url, std::string& data) {
+    // parse url
+    std::string host = parse_url(url);
+    // create connection
+    if (channels_.find(host) == channels_.end()) {
```
yq33victor: Do we need to hold all history connections ? Connections will consume local port and other resources. Maybe in the case of multiple different URL hosts, a large number of connections may be generated, consuming a large amount of local resources.

PR #105, opened 2025-09-08
refactor: refine gflags optimal value.

XuZhang99:


xllm/core/common/global_flags.cpp
```diff
@@ -87,69 +103,78 @@ DEFINE_double(max_memory_utilization,
...
-            true,
-            "Whether to enable schedule overlap.");
-
-DEFINE_double(prefill_scheduling_memory_usage_threshold,
-              0.95,
-              "The memory usage threshold during prefill scheduling.");
+DEFINE_string(
+    communication_backend,
+    "lccl",
+    "NPU communication backend.(e.g. lccl, hccl). When enable dp, use hccl.");
```
yq33victor, line 147: hccl only be used in dp ?

xllm/core/util/hash_util.h
```diff
@@ -63,10 +61,4 @@ struct FixedStringKeyEqual {
   }
 };
 
-void print_hex_array(uint8_t* array);
```
yq33victor, line 66: no need to move function `print_hex_array` to util.h, it  needs to use variable `MURMUR_HASH3_VALUE_LEN` here.

Actually, this function is not called and can be deleted.

xllm/core/common/global_flags.cpp
XuZhang99: when enable dp, hccl is better than lccl. on the contrary, using lccl is better.

xllm/core/util/hash_util.h
XuZhang99: this func is only used to debug for prefix cache and can be removed.

yq33victor (dismissed):
LGTM

xllm/core/common/global_flags.cpp
```diff
@@ -87,69 +103,78 @@ DEFINE_double(max_memory_utilization,
               "The fraction of GPU memory to be used for model inference, "
               "including model weights and kv cache.");
 
-DEFINE_int32(max_tokens_per_batch,
-             std::numeric_limits<int32_t>::max(),
-             "Max number of tokens per batch.");
+// --- scheduler config ---
+
+DEFINE_int32(max_tokens_per_batch, 20000, "Max number of tokens per batch.");
```
JimHsiung: Perhaps it would be better to set the default value to a power of 2.

xllm/core/common/global_flags.cpp
```diff
@@ -87,69 +103,78 @@ DEFINE_double(max_memory_utilization,
...
-DEFINE_int32(num_handling_threads, 4, "Number of handling threads.");
+DEFINE_bool(use_zero_evict,
+            false,
+            "Use ZeroEvictionScheduler but ContinuousScheduler.");
 
-DEFINE_int32(num_response_handling_threads,
-             4,
-             "Number of response handling threads.");
+DEFINE_int32(max_decode_token_per_sequence,
+             200,
```
JimHsiung: Same as above.

xllm/core/common/global_flags.cpp
```diff
@@ -15,26 +15,26 @@ limitations under the License.
...
 #include "brpc/reloadable_flags.h"
 
+// NOTE: related flags should be placed together.
+
 // --- xllm service config ---
 
 DEFINE_string(host, "", "Host name for brpc server.");
 
-DEFINE_int32(port, 8080, "Port for brpc server.");
+DEFINE_int32(port, 18000, "Port for brpc server.");
```
JimHsiung: It would be more appropriate to change the default port to 8010 to align with the online configuration.

xllm/core/common/global_flags.cpp
XuZhang99: max_tokens_per_batch is changed to 20480.

xllm/core/common/global_flags.cpp
XuZhang99: max_decode_token_per_sequence is changed to 256.

xllm/core/common/global_flags.cpp
XuZhang99: port is changed to 8010.

yq33victor (approved):

PR #108, opened 2025-09-08
feat: add profiler for latency prediction and initial latency-aware scheduler.

weizhehuang0827:
- Add profiling and time prediction modules for the continuous scheduler (mainly for chunkedprefill). 
- Initial process logic added into  build batch based on estimated latency.
- test scripts.
- token budget profile based on tpot for chunkedprefill (refer to Sarathi-Serve)

To add:
- seperate allocate prefix cache blocks and compute prefix cache match to estimate latency before acctually allocating blocks

### DIFF
--- a/xllm/core/common/global_flags.cpp
+++ b/xllm/core/common/global_flags.cpp
@@ -15,6 +15,8 @@ limitations under the License.
 
 #include "global_flags.h"
 
+#include <limits>
+
 #include "brpc/reloadable_flags.h"
 
 // NOTE: related flags should be placed together.
@@ -168,6 +170,10 @@ DEFINE_bool(enable_profile_step_time,
             false,
             "Whether to enable profile step time.");
 
+DEFINE_bool(enable_profile_token_budget,
+            false,
+            "Whether to enable profile token budget.");
+
 DEFINE_bool(enable_latency_aware_schedule,
             false,
             "use predicted latency for latency aware schedule.");
@@ -176,15 +182,15 @@ DEFINE_int32(profile_max_prompt_length,
              2048,
              "The max prompt length for profile.");
 
-DEFINE_bool(if_profile_kv_blocks,
+DEFINE_bool(enable_profile_kv_blocks,
             true,
             "true if generate kv cache for profile");
 
-DEFINE_int32(global_ttft_ms,
+DEFINE_int32(max_global_ttft_ms,
              std::numeric_limits<int32_t>::max(),
              "all requests use single global ttft");
 
-DEFINE_int32(global_tpot_ms,
+DEFINE_int32(max_global_tpot_ms,
              std::numeric_limits<int32_t>::max(),
              "all requests use single global ttft");
 
--- a/xllm/core/common/global_flags.cpp
+++ b/xllm/core/common/global_flags.cpp
@@ -162,6 +162,32 @@ DEFINE_string(rank_tablefile, "", "ATB HCCL rank table file.");
 
 DEFINE_int32(expert_parallel_degree, 0, "Expert parallel degree.");
 
+// --- profile config ---
+
+DEFINE_bool(enable_profile_step_time,
+            false,
+            "Whether to enable profile step time.");
+
+DEFINE_bool(enable_latency_aware_schedule,
+            false,
+            "use predicted latency for latency aware schedule.");
+
+DEFINE_int32(profile_max_prompt_length,
+             2048,
+             "The max prompt length for profile.");
+
+DEFINE_bool(if_profile_kv_blocks,
+            true,
+            "true if generate kv cache for profile");
+
+DEFINE_int32(global_ttft_ms,
+             std::numeric_limits<int32_t>::max(),
+             "all requests use single global ttft");
+
+DEFINE_int32(global_tpot_ms,
+             std::numeric_limits<int32_t>::max(),
+             "all requests use single global ttft");
+
 // --- prefix cache config ---
 
 DEFINE_bool(enable_prefix_cache,
--- a/xllm/core/scheduler/chunked_prefill_scheduler.cpp
+++ b/xllm/core/scheduler/chunked_prefill_scheduler.cpp
@@ -16,6 +16,8 @@ limitations under the License.
 
 #include "scheduler/chunked_prefill_scheduler.h"
 
+#include <limits>
+
 #include "common/metrics.h"
 #include "framework/batch/batch_factory.h"
 #include "util/timer.h"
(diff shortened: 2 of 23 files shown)


xllm/core/common/global_flags.cpp
```diff
@@ -139,6 +139,32 @@ DEFINE_bool(enable_mla,
...
+
+DEFINE_bool(enable_latency_aware_schedule,
+            false,
+            "use predicted latency for latency aware schedule.");
+
+DEFINE_int32(profile_max_prompt_length,
+             2048,
+             "The max prompt length for profile.");
+
+DEFINE_bool(if_profile_kv_blocks,
```
yq33victor: use `enable_` is better.

xllm/core/scheduler/chunked_prefill_scheduler.cpp
```diff
@@ -97,6 +105,25 @@ void ChunkedPrefillScheduler::handle_running_queue_requests(
         break;
       }
 
+      // for chunked prefill, we need to estimate latency according to
+      // current_step_handle_tokens TO IMPROVE and REFACTOR: seperate allocate
+      // prefix cache blocks and compute prefix cache match to estimate latency
+      // before acctually allocating blocks
+      // TODO:
+      size_t seq_estimate_latency = 0;
+      if (options_.enable_latency_aware_schedule()) {
```
yq33victor, line 111: in decode stage, every `Token` will predict time once, this may cost to much. Have we tested the predict time?

xllm/core/scheduler/chunked_prefill_scheduler.cpp
```diff
@@ -302,10 +340,33 @@ void ChunkedPrefillScheduler::handle_prefill_requests(
...
+          budget_exhausted = true;
+          break;
+        }
+      }
+
       prefill_sequences_budget.emplace_back(current_step_handle_tokens);
       prefill_sequences.emplace_back(prefill_sequence.get());
       allocated_tokens += current_step_handle_tokens;
       allocated_seqs += 1;
+      allocated_estimate_latency += seq_estimate_latency;
```
yq33victor, line 368: We shouldn't simply accumulate the forward time of each sequence?
For example, take two sequences a batch VS one sequence a batch, the computation time for both is most likely to be the same.

xllm/core/scheduler/continuous_scheduler.h
```diff
@@ -93,6 +94,18 @@ class ContinuousScheduler : public Scheduler {
...
+
+    PROPERTY(bool, enable_profile_step_time) = false;
+    // use predicted latency for latency aware schedule
+    PROPERTY(bool, enable_latency_aware_schedule) = false;
+    // the max prompt length for profile
+    PROPERTY(int32_t, profile_max_prompt_length) = 2048;
+    // true if generate kv cache for profile
+    PROPERTY(bool, if_profile_kv_blocks) = true;
+    // all requests use single global ttft
+    PROPERTY(int32_t, global_ttft_ms) = std::numeric_limits<int32_t>::max();
```
yq33victor: A similar name like `limited_global_ttft_ms/max_global_ttft_ms/...` might be a bit easier to understand. :)

global_ttft_ms, global_tpot_ms

xllm/core/scheduler/continuous_scheduler.cpp
```diff
@@ -48,10 +48,20 @@ ContinuousScheduler::ContinuousScheduler(Engine* engine, const Options& options)
...
 
   last_batch_.resize(options_.dp_size());
 
+  ProfileManager::Options profile_manager_options;
+  profile_manager_options.dp_size(options.dp_size())
+      .enable_schedule_overlap(options.enable_schedule_overlap())
+      .enable_profile_step_time(options.enable_profile_step_time())
+      .profile_max_prompt_length(options.profile_max_prompt_length())
+      .if_profile_kv_blocks(options.if_profile_kv_blocks());
+  profile_manager_ =
```
yq33victor: profile_manager_ is only needed when enable_latency_aware_schedule ?

xllm/core/scheduler/profile/time_predictor.cpp
```diff
@@ -0,0 +1,123 @@
...
+  // get coefficients
+  coefficients_ = matrix.colPivHouseholderQr().solve(target);
+  // output equation
+  LOG(INFO) << "Fitted equation: time = " << coefficients_(0) << " * diff^2 + "
+            << coefficients_(1) << " * diff + " << coefficients_(2)
+            << " * (diff * prefix_length) + " << coefficients_(3)
+            << " * prefix_length + " << coefficients_(4);
+}
+
+int32_t TimePredictor::predict_time(int32_t length, int32_t prefix_length) {
```
yq33victor: Is the accuracy of predict time of the algorithm high?

xllm/core/scheduler/profile/time_predictor.h
```diff
@@ -0,0 +1,41 @@
...
+#pragma once
+
+#include <Eigen/Dense>
+
+namespace xllm {
+
+// Predictor for predicting TTFT based on input length
+class TimePredictor final {
+ public:
+  TimePredictor(bool if_profile_prefix);
```
XuZhang99: add `explicit`

xllm/core/scheduler/chunked_prefill_scheduler.cpp
weizhehuang0827: We use simple linear model, consisting of only three additions and multiplications. I checked the `scheduling_latency` for safety. At 200*[2048,2048], there's almost no difference between using prediction (0.161s) and not using it (0.152s), and the latency is even lower with prediction enabled. Also, I remember we have `enable_schedule_overlap`?

xllm/core/scheduler/chunked_prefill_scheduler.cpp
weizhehuang0827: Thank you for pointing this out. I think I made an oversight. For seqs in a batch, the **constant term** in the linear model should not be repeatedly accumulated especially when q_seq_len is very short (e.g., for decode seq) bacause this term usually represents extra overhead for single forward. Only the term degree `>=1` should be accumulated.

xllm/core/scheduler/continuous_scheduler.cpp
weizhehuang0827: Yes. Maybe will need other profile like token budget profile and decode len prediction etc. in the future. I think I can add a judgement when Calling the constructor.

xllm/core/scheduler/profile/time_predictor.cpp
weizhehuang0827: I tested a single sequence input with a max length of 4000, a prompt step size of 40, and a kv cache token step size of 400. Currently, the MAE is about 8.4 when `profile_kv_blocks=false`, and about 3.8 when `profile_kv_blocks=true`.

xllm/core/scheduler/chunked_prefill_scheduler.cpp
weizhehuang0827: However, after testing, I found that the MAE without accumulating constant term  is much larger than the MAE with accumulating constant term. It may still be necessary to accumulate this term. orz

yq33victor (approved):
LGTM

liutongxuan (approved):

PR #110, opened 2025-09-08
feat: support multi_stream_parallel and modify batch sampler. [1/4]

guojinrong-nn:
Notice: this is the part-1/4 of multi_stream_parallel feature.

1. New added
- gflag : enable_multi_stream_parallel, default as false.
- add member variables `all_seqs_in_prefill_` in class `Batch`, to ensure the whole batch, all sequences in prefill or chunked prefill stage, to use this feature. If there is decode sequence, will not use this feature.
- add `concat` method in  class `SamplingParameters`, we split two micro batched to execute model execution individually, after this, will execute the sample operation together, to avoid extra small kernel launch.

2. Besides, refactor two places:
- delete useless function definition `split()` in `xllm/core/framework/batch/batch.h`.
- delete useless `set_dp_rank()` logic in `xllm/core/framework/batch/batch_factory.cpp`, dp_rank is already set in block manager.

### DIFF
--- a/xllm/core/common/global_flags.cpp
+++ b/xllm/core/common/global_flags.cpp
@@ -315,3 +315,11 @@ DEFINE_string(store_master_server_entry,
 DEFINE_string(store_metadata_connstring,
               "",
               "The address of the kv cache store metadata service.");
+
+// --- for computation communication parallel ---
+
+DEFINE_bool(
+    enable_multi_stream_parallel,
+    false,
+    "Whether to enable computation communication parallel by two streams "
+    "and two micro batches in prefill stage.");
--- a/xllm/core/framework/batch/batch_factory.cpp
+++ b/xllm/core/framework/batch/batch_factory.cpp
@@ -43,11 +43,10 @@ std::vector<Batch> BatchFactory::create_batches(
 
     // if dp enabled, each sequence is required to
     // dispatch to the same rank in the whole lifetime
-    if (sequence->dp_rank() >= 0) {
-      batches[sequence->dp_rank()].add(sequence, token_budget);
-    } else {
-      batches[i % dp_size_].add(sequence, token_budget);
-      sequence->set_dp_rank(i % dp_size_);
+    batches[sequence->dp_rank()].add(sequence, token_budget);
+    if (sequence->stage() == SequenceStage::DECODE &&
+        sequence->kv_state().kv_cache_tokens_num() > 0) {
+      batches[sequence->dp_rank()].set_batch_prefill_status(false);
     }
   }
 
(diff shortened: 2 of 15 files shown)


third_party/xllm_ops
LMX-xin, line 1: Submodule xllm_ops now uses a8e8b1, no need to fallback to 98b515

guojinrong-nn: updated.

xllm/core/common/global_flags.cpp
```diff
@@ -283,3 +283,11 @@ DEFINE_string(store_master_server_entry,
...
               "",
               "The address of the kv cache store metadata service.");
+
+// --- for computation communication parallel ---
+
+DEFINE_bool(
+    enable_multi_stream_parallel,
+    false,
+    "Whether to enable computation communication parallel by two streams "
+    "and two micro batches in prefill stage.");
```
yq33victor, line 325: only support prefill stage currently ?

xllm/core/util/tensor_helper.h
```diff
@@ -120,4 +120,14 @@ inline bool file_exists(const std::string& path) {
...
 
+inline torch::Tensor safe_concat(const torch::Tensor& t1,
+                                 const torch::Tensor& t2,
+                                 const uint32_t dim) {
+  if (t1.defined() && t2.defined()) {
+    return torch::cat({t1, t2}, dim);
+  } else if (!t2.defined()) {
+    return t1;
+  }
+}
```
yq33victor, line 132: the function will undefined when `t1 is undefined` and `t2 is defined`, need to return something back.

xllm/core/framework/batch/batch_factory.cpp
```diff
@@ -43,11 +43,10 @@ std::vector<Batch> BatchFactory::create_batches(
 
     // if dp enabled, each sequence is required to
     // dispatch to the same rank in the whole lifetime
-    if (sequence->dp_rank() >= 0) {
-      batches[sequence->dp_rank()].add(sequence, token_budget);
-    } else {
-      batches[i % dp_size_].add(sequence, token_budget);
-      sequence->set_dp_rank(i % dp_size_);
+    batches[sequence->dp_rank()].add(sequence, token_budget);
```
yq33victor, line 46: when to assign dp_rank to sequence ?

xllm/core/common/global_flags.cpp
guojinrong-nn: Yes, decode stage is not supported temporarily.

xllm/core/framework/batch/batch_factory.cpp
guojinrong-nn: It's assigned in BlockManagerPool currently, happens when allocating kv cache for the sequence.

xllm/core/util/tensor_helper.h
guojinrong-nn: fixed.

PR #114, opened 2025-09-10
feat: add image generation API.

yiming-l21:
feat: add image generation API.

### DIFF
--- a/xllm/core/framework/request/dit_request_params.h
+++ b/xllm/core/framework/request/dit_request_params.h
@@ -1,0 +1,60 @@
+#pragma once
+#include <cstdint>
+#include <nlohmann/json.hpp>
+#include <optional>
+#include <string>
+#include <vector>
+
+#include "tensor.pb.h"
+
+namespace xllm {
+struct GenerationParams {
+  std::optional<std::string> size;
+
+  int32_t width = 512;
+
+  int32_t height = 512;
+
+  std::optional<int32_t> num_inference_steps;
+
+  std::optional<float> true_cfg_scale;
+
+  std::optional<float> guidance_scale;
+
+  std::optional<uint32_t> num_images_per_prompt = 1;
+
+  std::optional<int64_t> seed;
+
+  std::optional<int32_t> max_sequence_length;
+};
+
+struct InputParams {
+  std::string prompt;
+
+  std::optional<std::string> prompt_2;
+
+  std::optional<std::string> negative_prompt;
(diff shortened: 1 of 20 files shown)


xllm/core/framework/request/request_output.cpp
```diff
@@ -51,4 +51,34 @@ void RequestOutput::log_request_status() const {
   }
 }
 
+void ImageRequestOutput::log_request_status() const {
+  if (!status.has_value()) {
```
yq33victor: nit: `log_request_status` -> `count_request_status` ?

xllm/proto/tensor.proto
```diff
@@ -0,0 +1,23 @@
...
+message TensorContents
+{
+  repeated bool bool_contents = 1;
+  repeated int32 int_contents = 2;
+  repeated int64 int64_contents = 3;
+  repeated uint32 uint_contents = 4;
+  repeated uint64 uint64_contents = 5;
+  repeated float fp32_contents = 6;
+  repeated double fp64_contents = 7;
+  repeated bytes bytes_contents = 8;
```
yq33victor, line 15: nit: Is it enough to use `repeated bytes`? 

Anyway, there's no problem with the current struct.

xllm/core/framework/request/request_output.h
```diff
@@ -76,6 +76,22 @@ struct SequenceOutput {
   // the embeddings of the prompt token
   std::optional<std::vector<float>> embeddings;
 };
+struct ImageGenerationOutput {
+  // the index of the sequence in the request.
+  size_t index;
+
+  // the generated image in proto tensor format.
+  proto::Tensor image_tensor;
```
yq33victor: We'd better not use `Protobuf type` at runtime. We can use `torch::Tensor` here instead.

xllm/core/framework/request/dit_request_params.h
```diff
@@ -0,0 +1,60 @@
...
+
+  std::optional<std::string> negative_prompt_2;
+
+  // std::optional<std::string> ip_adapter_image;
+
+  // std::optional<std::string> negative_ip_adapter_image;
+
+  std::optional<proto::Tensor> prompt_embeds;
+
+  std::optional<proto::Tensor> pooled_prompt_embeds;
```
yq33victor: nit: `proto::Tensor` -> `torch::Tensor` is better ?

PR #128, opened 2025-09-12
feat: support multi_stream_parallel and modify engine to split micro batches. [2/4]

guojinrong-nn:
1. Mainly support engine to split micro batches for multi_stream_parallel feature.

2. Beside, the most of changes is LLMEngine is introduced by abstracting three frequently used args as class member variable, to avoid duplicated computation.

-   uint32_t dp_size_;
-   uint32_t worker_clients_num_;
-   uint32_t dp_local_tp_size_;

### DIFF
--- a/xllm/core/runtime/llm_engine.cpp
+++ b/xllm/core/runtime/llm_engine.cpp
@@ -655,16 +655,15 @@ void LLMEngine::process_eplb_data(
   eplb_manager_->update_expert_load(tensors);
 }
 
-uint32_t determine_micro_batches_num(std::vector<Batch>& batch) {
+uint32_t determine_micro_batches_num(const std::vector<Batch>& batch) {
   bool is_all_prefill =
       std::all_of(batch.begin(), batch.end(), [](const Batch& one_batch) {
         return one_batch.get_batch_prefill_status();
       });
-  if (!is_all_prefill ||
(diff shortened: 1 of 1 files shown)


xllm/core/runtime/llm_engine.cpp
```diff
@@ -676,4 +655,68 @@ void LLMEngine::process_eplb_data(
...
+      std::all_of(batch.begin(), batch.end(), [](const Batch& one_batch) {
+        return one_batch.get_batch_prefill_status();
+      });
+  if (!is_all_prefill ||
+      (is_all_prefill && !FLAGS_enable_multi_stream_parallel)) {
+    return 1;
+  } else if (FLAGS_enable_multi_stream_parallel) {
+    return 2;
+  }
+}
```
yq33victor, line 668: emmm :) , else { return ... }

guojinrong-nn: anything else? gonna fix together.

yq33victor: > anything else? gonna fix together.

just fix it! :)

guojinrong-nn: done

liutongxuan (approved):

PR #135, opened 2025-09-16
feat: support beam search for llm model [1/N].

RobbieLeung:
## Background

basic beam search support for llm model and generative recommendation.

## Usage

```

curl http://127.0.0.1:13222/v1/completions
	-H 'accept: application/json' -H 'Content-Type: application/json'
	-X 'POST'
	-d '{
		"model": "Qwen3-8B",
		"prompt": "hello xllm",
		"stream": false,
		"max_tokens": 30,
		"temperature": 0.0,
		"beam_width": 5,
		"logprobs": 5
  	}'

```

## More Work

* copy blocks kernel.
* optimize cpu cost of beam search.
* beam search kernel.
* paged attention kernel for beam search.

### DIFF
--- a/xllm/core/framework/batch/batch.cpp
+++ b/xllm/core/framework/batch/batch.cpp
@@ -75,6 +75,7 @@ ForwardInput Batch::prepare_forward_input(uint32_t num_decoding_tokens,
                             mm_data_vec_,
                             copy_in_cache_block_infos_,
                             copy_out_cache_block_infos_,
+                            swap_cache_block_infos_,
                             &args);
   return builder.build_forward_input(num_decoding_tokens,
                                      min_decoding_batch_size);
@@ -88,6 +89,7 @@ RawForwardInput Batch::prepare_forward_input(uint32_t start_idx,
                             mm_data_vec_,
                             copy_in_cache_block_infos_,
                             copy_out_cache_block_infos_,
+                            swap_cache_block_infos_,
                             nullptr);
   return builder.build_raw_forward_input(start_idx, end_idx);
 }
@@ -134,6 +136,7 @@ void Batch::process_sample_output(const RawForwardOutput& raw_output,
     }
   }
   CHECK_EQ(output_idx, num_seqs);
+  process_beam_search();
 }
 
 void Batch::process_sample_output(const SampleOutput& sample_output,
@@ -175,6 +178,7 @@ void Batch::process_sample_output(const SampleOutput& sample_output,
     append_token_for_sequence(seq, token, 0, enable_schedule_overlap);
   }
   CHECK_EQ(output_idx, num_seqs);
+  process_beam_search();
 }
 
 bool Batch::update_sequence_state(Sequence* seq, bool enable_schedule_overlap) {
@@ -244,6 +248,12 @@ void Batch::process_embedding_output(const torch::Tensor& output_embedding) {
       seq->update_embeddings(seq_img_embedding);
       seq->append_token(token);
     }
+  }
+}
+
+void Batch::process_beam_search() {
+  for (auto* sequence_group : sequence_groups_) {
+    sequence_group->process_beam_search();
   }
 }
 }  // namespace xllm
--- a/xllm/core/framework/batch/batch_factory.cpp
+++ b/xllm/core/framework/batch/batch_factory.cpp
@@ -17,11 +17,25 @@ limitations under the License.
 
 namespace xllm {
 
+namespace {
+
+bool is_beam_search(const std::vector<std::shared_ptr<Request>>& requests) {
+  for (const auto& request : requests) {
+    if (request->check_beam_search()) {
+      return true;
+    }
+  }
+  return false;
+}
+}  // namespace
+
 std::vector<Batch> BatchFactory::create_batches(
+    const std::vector<std::shared_ptr<Request>>& running_requests,
     const std::vector<Sequence*>& running_sequences,
     const std::vector<size_t>& running_sequences_budgets,
     std::vector<std::vector<CacheBlockInfo>>* copy_in_cache_block_infos,
-    std::vector<std::vector<CacheBlockInfo>>* copy_out_cache_block_infos) {
+    std::vector<std::vector<CacheBlockInfo>>* copy_out_cache_block_infos,
+    std::vector<std::vector<CacheBlockInfo>>* swap_cache_block_infos) {
   size_t num_prompt_tokens = 0;
   size_t num_generated_tokens = 0;
   std::vector<Batch> batches(dp_size_);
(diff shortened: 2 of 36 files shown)


xllm/core/framework/batch/batch_factory.cpp
```diff
@@ -50,6 +64,14 @@ std::vector<Batch> BatchFactory::create_batches(
     }
   }
 
+  if (is_beam_search(running_requests)) {
+    for (const auto& request : running_requests) {
+      auto seq_group = request->sequence_group();
+      int32_t dp_rank = seq_group->dp_rank();
+      batches[dp_rank].add(seq_group);
```
yq33victor, line 71: nit: could we add a function named `int32_t get_dp_rank() {...}` in SequenceGroup, that we don't need to return raw pointer from unique_ptr<SequenceGroup>. :)

```
auto seq_group = request->sequence_group();
int32_t dp_rank = seq_group->dp_rank();
```
--->
```
int32_t dp_rank = seq_group->get_dp_rank();
```

xllm/core/framework/batch/beam_search.h
```diff
@@ -0,0 +1,116 @@
+#pragma once
+
```
yq33victor, line 17: nit: add copyright header 
```
/* Copyright 2025 The xLLM Authors. All Rights Reserved.
....
```

xllm/core/framework/batch/beam_search.h
```diff
@@ -0,0 +1,116 @@
...
+  BeamCandidate() = default;
+
+  BeamCandidate(size_t seq_idx,
+                float logprob,
+                std::vector<int32_t>& token_ids,
+                std::vector<std::optional<float>>& logprobs)
+      : seq_index(seq_idx),
+        logprob_sum(logprob),
+        token_ids(token_ids),
```
yq33victor: nit:
```
token_ids(std::move(token_ids)),
logprobs(std::move(logprobs)) {}
```

xllm/core/framework/batch/beam_search.h
```diff
@@ -0,0 +1,116 @@
...
+  void insert(CandidateType&& candidate) {
+    if (min_heap_.size() < k_) {
+      min_heap_.push(std::move(candidate));
+    } else if (candidate.logprob_sum > min_heap_.top().logprob_sum) {
+      min_heap_.pop();
+      min_heap_.push(std::move(candidate));
+    }
+  }
+
+  void insertBatch(const std::vector<CandidateType>& candidates) {
```
yq33victor: nit: function name, `insertBatch` -> `insert_batch`

and others functions below.

xllm/core/framework/batch/beam_search.h
```diff
@@ -0,0 +1,116 @@
...
+      result.emplace_back(
+          std::move(const_cast<CandidateType&>(min_heap_.top())));
+      min_heap_.pop();
+    }
+
+    return result;
+  }
+
+  std::vector<CandidateType>&& getTopKMove() {
+    static thread_local std::vector<CandidateType> result;
```
yq33victor: nit: emmm, why we need static thread_local here, result is seems be cleared evetytime :(
And it will be invalid after move out at last.

xllm/core/framework/request/sequence.cpp
```diff
@@ -65,6 +65,39 @@ Sequence::Sequence(size_t index,
   cur_generated_token_idx_ = num_prompt_tokens_;
 }
 
+// 实现拷贝构造逻辑
+Sequence::Sequence(const Sequence& other)
```
yq33victor, line 68: nit: english comments 🥇

xllm/core/framework/request/sequence.cpp
```diff
@@ -65,6 +65,39 @@ Sequence::Sequence(size_t index,
   cur_generated_token_idx_ = num_prompt_tokens_;
 }
 
+// 实现拷贝构造逻辑
```
liutongxuan: comments should be written in english.

xllm/core/framework/batch/batch.cpp
```diff
@@ -133,6 +135,7 @@ void Batch::process_sample_output(const RawForwardOutput& raw_output,
     }
   }
   CHECK_EQ(output_idx, num_seqs);
+  process_beam_search();
```
liutongxuan, line 139: Is it necessary to add a new class of batch for beam search?

xllm/core/framework/request/sequences_group.cpp
```diff
@@ -113,4 +122,148 @@ void SequencesGroup::generate_outputs(std::vector<SequenceOutput>& outputs,
...
+    const auto& log_probs = seq->logprob_state()->get_logprobs();
+    const auto& top_logprobs =
+        seq->logprob_state()->get_top_logprobs()[last_token_idx];
+    const auto& top_tokens =
+        seq->logprob_state()->get_top_tokens()[last_token_idx];
+    float base_logprob =
+        seq->get_average_logprob() * num_generated_tokens - top_logprobs[0];
+
+    if (seq_size == 1) {
+      direct_candidates.reserve(beam_width);
```
yq33victor: nit: it's works fine.
but, no need the `if` here ? we can use `topk_optimizer` when seq_size == 1 also. The time complexity of the function `priority_queue.insert(...)` is O(logN).

xllm/core/framework/block/block_manager_pool.cpp
```diff
@@ -218,6 +228,29 @@ std::vector<Block> BlockManagerPool::allocate(size_t num_tokens,
...
+  auto src_blocks = sequence->kv_state().src_blocks();
+  if (src_blocks.size() == 0) {
+    return;
+  }
+
+  // when sequence need to swap the last block and no new block appended,
+  // allocate a new block for this sequence
+  if (need_swap && sequence->kv_state().need_swap()) {
+    int32_t dp_rank = get_dp_rank(sequence);
+    auto new_blocks = block_managers_[dp_rank]->allocate(1);
```
yq33victor, line 245: What happens if block cannot be allocated here?

RobbieLeung: If a parent sequence has multiple selected sub-sequences, only the first sub-sequence can reuse the parent’s last block. The others must allocate new blocks and copy data from the parent.

xllm/core/framework/request/sequences_group.cpp
RobbieLeung: The time complexity of topk_optimizer is O(N log N) because log_probs are pre-sorted when seq_size = 1. However, this logic should be moved to the frontend part. The current implementation contains a bug which I will fix.

xllm/core/framework/block/block_manager_pool.cpp
RobbieLeung: If a parent sequence has multiple selected sub-sequences, only the first sub-sequence can reuse the parent’s last block. The others must allocate new blocks and copy data from the parent.

RobbieLeung: If a parent sequence has multiple selected sub-sequences, only the first sub-sequence can reuse the parent’s last block. The others must allocate new blocks and copy data from the parent.

xllm/core/framework/batch/batch.cpp
RobbieLeung: The beam search processing occurs within `SequencesGroup`, while `Batch` is only responsible for triggering its execution. In a future refactoring, the triggering logic will be moved to the `Scheduler`, which will then schedule requets based on `SequencesGroup`. Additionally, partial scheduling of `Sequences` may influence the beam search process.

PR #138, opened 2025-09-17
feat: support multi_stream_parallel and modify all workers. [3/4]

guojinrong-nn:
Main changes include:

- change `ForwardInput` to `BatchedForwardInputs` in `step` function across various workers.

### DIFF
--- a/xllm/core/runtime/llm_worker_impl.cpp
+++ b/xllm/core/runtime/llm_worker_impl.cpp
@@ -78,47 +78,63 @@ bool LLMWorkerImpl::init_model(ModelContext& context) {
   return true;
 }
 
-std::optional<ForwardOutput> LLMWorkerImpl::step(const ForwardInput& inputs) {
+std::optional<ForwardOutput> LLMWorkerImpl::step(
+    const BatchedForwardInputs& inputs) {
 #if defined(USE_NPU)
   c10_npu::SetDevice(device_.index());
 #elif defined(USE_MLU)
   // TODO(mlu): implement mlu set device
 #endif
   Timer timer;
-  auto& flatten_tokens = inputs.token_ids;
-  auto& flatten_positions = inputs.positions;
-  auto& params = inputs.input_params;
-  auto& sampling_params = inputs.sampling_params;
+  std::vector<torch::Tensor> flatten_tokens_micro_batches;
+  std::vector<torch::Tensor> flatten_positions_micro_batches;
+  std::vector<ModelInputParams> input_params_micro_batches;
+  auto& concated_sampling_params = inputs.concated_sampling_params;
 
-  if (FLAGS_enable_eplb) {
-    eplb_executor_->eplb_execute(inputs.eplb_info);
-  }
   std::vector<folly::SemiFuture<bool>> futures;
-  if (options_.instance_role() == InstanceRole::PREFILL &&
-      options_.kv_cache_transfer_mode() == "PUSH" &&
-      !inputs.transfer_kv_infos.empty()) {
-#if defined(USE_NPU)
-    std::shared_ptr<NPULayerSynchronizerImpl> layer_synchronizer =
-        std::make_shared<NPULayerSynchronizerImpl>(
-            context_.get_model_args().n_layers());
-    const_cast<ModelInputParams*>(&params)->layer_synchronizer =
-        layer_synchronizer;
 
-    futures.emplace_back(
-        kv_cache_transfer_->push_kv_blocks_async(inputs.transfer_kv_infos,
-                                                 context_.get_parallel_args(),
-                                                 layer_synchronizer,
-                                                 is_spec_draft_));
+  for (auto i = 0; i < inputs.micro_inputs.size(); ++i) {
+    flatten_tokens_micro_batches.push_back(
+        std::move(inputs.micro_inputs[i].token_ids));
+    flatten_positions_micro_batches.push_back(
+        std::move(inputs.micro_inputs[i].positions));
+    input_params_micro_batches.push_back(
+        std::move(inputs.micro_inputs[i].input_params));
+
+    if (options_.instance_role() == InstanceRole::PREFILL &&
+        options_.kv_cache_transfer_mode() == "PUSH" &&
+        !inputs.micro_inputs[i].transfer_kv_infos.empty()) {
+#if defined(USE_NPU)
+      std::shared_ptr<NPULayerSynchronizerImpl> layer_synchronizer =
+          std::make_shared<NPULayerSynchronizerImpl>(
+              context_.get_model_args().n_layers());
+      const_cast<ModelInputParams*>(&(input_params_micro_batches[i]))
+          ->layer_synchronizer = layer_synchronizer;
+
+      futures.emplace_back(kv_cache_transfer_->push_kv_blocks_async(
+          inputs.micro_inputs[i].transfer_kv_infos,
+          context_.get_parallel_args(),
+          layer_synchronizer,
+          is_spec_draft_));
 #endif
+    }
+  }
+  if (FLAGS_enable_eplb) {
+    eplb_executor_->eplb_execute(inputs.micro_inputs[0].eplb_info);
   }
 
+  // temporarily use [0], will be adapted in next pr
   // call model executor forward to get hidden states
-  auto hidden_states = model_executor_->forward(
-      flatten_tokens, flatten_positions, kv_caches_, params);
+  auto hidden_states =
+      model_executor_->forward(flatten_tokens_micro_batches[0],
+                               flatten_positions_micro_batches[0],
+                               kv_caches_,
+                               input_params_micro_batches[0]);
+
   torch::Tensor logits;
-  if (sampling_params.selected_token_idxes.defined()) {
-    logits =
-        model_->logits(hidden_states, sampling_params.selected_token_idxes);
+  if (concated_sampling_params.selected_token_idxes.defined()) {
+    logits = model_->logits(hidden_states,
+                            concated_sampling_params.selected_token_idxes);
   }
 
   ForwardOutput output;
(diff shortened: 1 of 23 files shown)


xllm/core/runtime/llm_worker_impl.cpp
```diff
@@ -85,47 +85,64 @@ bool LLMWorkerImpl::init_model(torch::ScalarType dtype,
   return true;
 }
 
-std::optional<ForwardOutput> LLMWorkerImpl::step(const ForwardInput& inputs) {
+std::optional<ForwardOutput> LLMWorkerImpl::step(
```
liutongxuan, line 81: We should add new interfaces to handle BatchForwardInput instead of modifying the logic in the ForwardInput interface？

guojinrong-nn: If add a new interface, all functions along the whole `engine-worker-executer-model` path will be implemented twice, and the BatchForwardInput only involves vector dimension, nothing else, ForwardInput case can be coverd completely.  or introduce a template?
Any advices?@yq33victor

xllm/core/runtime/llm_worker_impl.cpp
```diff
@@ -85,47 +85,64 @@ bool LLMWorkerImpl::init_model(torch::ScalarType dtype,
...
+    const BatchedForwardInputs& inputs) {
 #if defined(USE_NPU)
```
yq33victor, line 83: I noticed that there are some duplicate definitions in `struct BatchedForwardInputs`. Currently, it contain a `std::vector<ForwardInput>` variable, but the `ModelInputParams, transfer_kv_infos, and eplb_info` in `struct ForwardInput` don't need to be stored in each micro input. All inputs share these variables, so `struct BatchedForwardInputs` and `struct ForwardInput` may need some optimization.

nit: If there's no time to modify it in this PR, a new PR can be submitted later to make the optimizations.

xllm/core/runtime/llm_worker_impl.cpp
```diff
@@ -85,47 +85,64 @@ bool LLMWorkerImpl::init_model(torch::ScalarType dtype,
...
+      futures.emplace_back(kv_cache_transfer_->push_kv_blocks_async(
+          inputs.micro_inputs[i].transfer_kv_infos,
+          context_.get_parallel_args(),
+          layer_synchronizer,
+          is_spec_draft_));
 #endif
+    }
+
+    if (FLAGS_enable_eplb) {
+      eplb_executor_->eplb_execute(inputs.micro_inputs[i].eplb_info);
```
yq33victor: There is no need to execute `eplb_execute` multiple times, and this is not related to the number of micro_inputs.

@DongheJin  please confirm it, thanks.

xllm/core/runtime/llm_worker_impl.cpp
```diff
@@ -85,47 +85,64 @@ bool LLMWorkerImpl::init_model(torch::ScalarType dtype,
...
-    const_cast<ModelInputParams*>(&params)->layer_synchronizer =
-        layer_synchronizer;
-
-    futures.emplace_back(
-        kv_cache_transfer_->push_kv_blocks_async(inputs.transfer_kv_infos,
-                                                 context_.get_parallel_args(),
-                                                 layer_synchronizer,
-                                                 is_spec_draft_));
+      std::shared_ptr<NPULayerSynchronizerImpl> layer_synchronizer =
+          std::make_shared<NPULayerSynchronizerImpl>(
```
yq33victor, line 109: There is no need to execute` push_kv_blocks_async` multiple times, and this is not related to the number of micro_inputs.

@JimHsiung please confirm it, thanks.

xllm/core/runtime/llm_worker_impl.cpp
yq33victor: > If add a new interface, all functions along the whole `engine-worker-executer-model` path will be implemented twice, and the BatchForwardInput only involves vector dimension, nothing else, ForwardInput case can be coverd completely. or introduce a template? Any advices?@yq33victor

emmm, can we call `step(ForwardInput& inputs)` in function `step(const BatchedForwardInputs& inputs)` ? This will make the entire code look clearer. However, I don't know what the specific multi-stream parallel code looks like, so it's uncertain whether it can be easily refactored into this form.
```
std::optional<ForwardOutput> LLMWorkerImpl::step(const BatchedForwardInputs& inputs) {
  // do something 
  for (auto& input: inputs) {
    step(input);
  }
  // do something
  sync.
  return output;
}
```

nit:
And maybe we don't named `BatchedForwardInputs` type, a general name is better, like `ForwardInputs`, and we define micro inputs in the type, case that we can handle some shared params in `ForwardInputs`, and no need to params copy to all micro inputs.

xllm/core/runtime/worker_impl.cpp
```diff
@@ -459,25 +469,39 @@ folly::SemiFuture<bool> WorkerImpl::copy_out_blocks_async(
...
+  BatchedForwardInputs batched_inputs_on_device;
+  batched_inputs_on_device.micro_inputs.reserve(inputs.micro_inputs.size());
+
+  prepare_work_before_execute(inputs, batched_inputs_on_device);
 
   folly::Promise<std::optional<ForwardOutput>> promise;
   auto future = promise.getSemiFuture();
   threadpool_.schedule([this,
-                        inputs = forward_inputs_on_device,
+                        inputs = batched_inputs_on_device,
```
yq33victor: nit: inputs = std::move(batched_inputs_on_device),

xllm/core/runtime/speculative_worker_impl.cpp
```diff
@@ -168,21 +168,23 @@ bool SpeculativeWorkerImpl::allocate_kv_cache_with_transfer(
 #endif
 
 std::optional<ForwardOutput> SpeculativeWorkerImpl::step(
-    const ForwardInput& inputs) {
-  if (inputs.token_ids.numel() == 0) {
+    const BatchedForwardInputs& inputs) {
+  // all micro batches in multi stream parallel share the same
+  // prefill/decode stage, use inputs[0] here
+  if (inputs.micro_inputs[0].token_ids.numel() == 0) {
```
yq33victor, line 174: Is it possible that some micro inputs are empty ? Likee there is just one request.

And how to handle this case.

xllm/core/runtime/speculative_worker_impl.cpp
```diff
@@ -230,78 +248,104 @@ std::optional<ForwardOutput> SpeculativeWorkerImpl::step_prefill(
...
-      !inputs.transfer_kv_infos.empty()) {
-    auto future = kv_cache_transfer_->push_kv_blocks_async(
-        inputs.transfer_kv_infos, context_.get_parallel_args(), nullptr, true);
-    auto out = std::move(future).get();
+  for (auto i = 0; i < inputs.micro_inputs.size(); ++i) {
+    if (options_.instance_role() == InstanceRole::PREFILL &&
+        options_.kv_cache_transfer_mode() == "PUSH" &&
+        !inputs.micro_inputs[i].transfer_kv_infos.empty()) {
+      auto future = kv_cache_transfer_->push_kv_blocks_async(
+          inputs.micro_inputs[i].transfer_kv_infos,
```
yq33victor, line 278: no need to handle multi times ？ @JimHsiung  please help to confirm it.

xllm/core/runtime/speculative_worker_impl.cpp
```diff
@@ -230,78 +248,104 @@ std::optional<ForwardOutput> SpeculativeWorkerImpl::step_prefill(
...
+    auto& input = inputs.micro_inputs[i];
+    auto& draft_input = draft_inputs.micro_inputs[i];
+    // get embedding cache
 #if defined(USE_NPU)
-  if (kv_cache_transfer_) {
-    kv_cache_transfer_->copy_blocks(inputs.input_params.embedding_ids,
-                                    /*h2d*/ false);
-  }
+    if (kv_cache_transfer_) {
+      kv_cache_transfer_->copy_blocks(input.input_params.embedding_ids,
```
yq33victor, line 336: here is same to above. @JimHsiung please help to confirm it.

xllm/core/runtime/speculative_worker_impl.cpp
```diff
@@ -345,20 +409,24 @@ std::optional<ForwardOutput> SpeculativeWorkerImpl::step_decode(
...
+    embedding_allocator_->write_validate(input.input_params.embedding_ids,
+                                         val_output.next_tokens.to(torch::kCPU),
+                                         val_output.embeddings);
 #if defined(USE_NPU)
-  if (kv_cache_transfer_) {
-    kv_cache_transfer_->copy_blocks(inputs.input_params.embedding_ids,
-                                    /*h2d*/ true);
-  }
+    if (kv_cache_transfer_) {
+      kv_cache_transfer_->copy_blocks(input.input_params.embedding_ids,
```
yq33victor, line 424: here @JimHsiung

xllm/core/runtime/llm_worker_impl.cpp
DongheJin: Confirmed - eplb_execute should only be executed once, independent of micro_inputs count.

guojinrong-nn: already moved out of the for loop

xllm/core/runtime/llm_worker_impl.cpp
guojinrong-nn: ok, I have started to refine the code as suggested, just block this pr until the later update

xllm/core/runtime/llm_worker_impl.cpp
guojinrong-nn: Most of the ModelInputParams, SamplingParameters, and transfer_kv_infos are related to the sequences of each micro batches, except eplb_info shares the same one.

xllm/core/runtime/speculative_worker_impl.cpp
guojinrong-nn: this is for dp, a certain dp rank may be empty.

xllm/core/runtime/worker_impl.cpp
guojinrong-nn: fixed

yq33victor (approved):

liutongxuan (approved):

PR #141, opened 2025-09-17
refactor: simplify atb context and workspace call chain.

RobbieLeung:


xllm/core/framework/model_context.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+#include <torch/torch.h>
+#ifdef TORCH_HIGHER_THAN_PTA6
+// #include <torch_npu/csrc/core/npu/NPUFormat.h>
+#include <torch_npu/csrc/framework/OpCommand.h>
+#else
+#include <torch_npu/csrc/aten/NPUNativeFunctions.h>
+#include <torch_npu/csrc/framework/utils/OpPreparation.h>
+#endif
+#include <acl/acl.h>
```
JimHsiung: A macro needs to be added here to distinguish between NPU and MLU.

xllm/core/framework/model_context.h
```diff
@@ -15,6 +15,8 @@ limitations under the License.
 
 #pragma once
 
+#include <acl/acl.h>
```
JimHsiung, line 19: Same as above.

xllm/core/layers/npu/atb_base.cpp
```diff
@@ -208,16 +209,17 @@ atb::Status ATBBase::execute_node(atb_speed::Model::Node& node,
        << std::endl;
     throw std::runtime_error(ss.str());
   }
-  context_ = context;
-  atb::Status st =
-      node.operation->Setup(node.variantPack, node.workspaceSize, context_);
+
+  atb::Status st = node.operation->Setup(node.variantPack,
+                                         node.workspaceSize,
+                                         const_cast<atb::Context*>(context_));
```
JimHsiung: Perhaps converting the member variable to a non-const type would be better.

xllm/core/runtime/vlm_worker_impl.cpp
```diff
@@ -66,19 +64,11 @@ bool VLMWorkerImpl::init_model(torch::ScalarType dtype,
...
-  tmp_model_args.image_embedding_mode() = false;
-  context_.set_model_args(tmp_model_args);
-
-  context_.set_quant_args(quant_args);
-
-  dtype_ = dtype;
-  context_.set_tensor_options(torch::dtype(dtype_).device(device_));
-
-  model_ = create_vlm_model(context_);
+  context.set_image_embedding_mode(true);
```
JimHsiung: true -> false

PR #176, opened 2025-09-24
feat: acclerate generate output by using multi thread.

RobbieLeung:


xllm/core/scheduler/async_response_processor.h
```diff
@@ -64,6 +64,9 @@ class AsyncResponseProcessor final {
   // the threadpool to handle rpc
   ThreadPool rpc_threadpool_;
 
+  // the threadpool to generate outputs
+  ThreadPool generate_output_threadpool_{16};
```
yq33victor, line 68: could we move `generate_output_threadpool_` to Request ?

we can pass a `bool` flag.

RobbieLeung: `Request` will be created and destroyed frequently. `AsyncResponseProcessor`  has the whole lifestyle with program.

liutongxuan (approved):

PR #219, opened 2025-10-10
feat: add unified management for stream.

XuZhang99:


xllm/core/platform/stream_helper.h
```diff
@@ -0,0 +1,41 @@
...
+
+#include <cstdint>
+#if defined(USE_NPU)
+#include <torch_npu/csrc/framework/OpCommand.h>
+#include <torch_npu/torch_npu.h>
+#endif
+
+namespace xllm {
+
+void synchronize_stream(int32_t device_id);
```
yq33victor: nit: use a namespace like `stream` for global function? just a suggestion.

xllm/core/platform/stream_helper.cpp
```diff
@@ -0,0 +1,39 @@
...
+#endif
+
+void StreamHelper::synchronize_stream() {
+#if defined(USE_NPU)
+  aclrtSynchronizeStream(stream_.stream());
+#elif defined(USE_MLU)
+  // TODO(mlu): implement mlu synchronize stream
+#endif
+}
+
```
yq33victor, line 18: maybe we can delete `synchronize_stream_return_status`. 
use `int StreamHelper::synchronize_stream()` is enough? user can decide if to handle the `returned status` or not.

xllm/core/platform/stream_helper.h
liutongxuan: It should not exist as an independent function. If the stream of the device needs to be synchronized, it should still follow the processing logic of the unified Stream class and not be an independent function

XuZhang99: `synchronize_stream(int32_t device_id)` is a static function in `StreamHelper` now.

XuZhang99: `synchronize_stream(int32_t device_id)` is a static function in `StreamHelper` now.

xllm/core/platform/stream_helper.cpp
XuZhang99: both `synchronize_stream()` and `synchronize_stream(int32_t device_id)` already have return status.

PR #226, opened 2025-10-13
 refactor: rename StreamHelper to Stream and update stream management.

XuZhang99:


xllm/core/platform/device.cpp
```diff
@@ -74,11 +67,35 @@ int64_t Device::total_memory() { return get_device_mem().total_memory; }
...
-int Device::synchronize_stream() {
+int Device::synchronize_main_stream() {
 #if defined(USE_NPU)
   return aclrtSynchronizeStream(c10_npu::getCurrentNPUStream(index()).stream());
 #elif defined(USE_MLU)
   // TODO(mlu): implement mlu synchronize stream
 #endif
 }
+
+void Device::get_stream_from_pool(const std::string& stream_name) {
```
liutongxuan: 是不是应该增加一个get_default_stream? 看调用的地方都是“sample_stream“这个是个default么？

XuZhang99: this function retrieves a stream from the stream pool and assigns a name to the stream. sometimes we need to get more than one stream form stream pool, so we needs a stream name to distinguish different stream.

PR #242, opened 2025-10-16
bugfix: resolve segmentation fault during qwen3 quantized inference.

yingxudeng:


yingxudeng:
Due to the incompatibility in interfaces and data types between the RMSNorm (5 inputs, 1 output, int8) required for quantization and aclnnRmsNorm (2 inputs, 2 outputs, bf16), I modified the ATB graph by adding Cast operators for type conversion. Although this change prevented crashes, it resulted in garbled output. Since this optimization primarily targets the prefill phase, we have temporarily disabled the invocation of aclnnRmsNorm during quantization for now.

JimHsiung (approved):

PR #269, opened 2025-10-23
feat: support beam search kernel for decode-only model.

yjw258:
Use beam search kernel to select candidates in beam search kernel, and add a launch parameter `enable_beam_search_kernel` to control this behavior.

### DIFF
--- a/xllm/core/distributed_runtime/worker_service.cpp
+++ b/xllm/core/distributed_runtime/worker_service.cpp
@@ -346,6 +346,20 @@ void WorkerService::ExecuteModel(
           batched_fwd_inputs.micro_inputs[i].sampling_params);
     }
 
+    // concat acc_logprob here for beam search together
+    if (micro_batches_num > 1) {
+      std::vector<torch::Tensor> acc_logprob_vec;
+      acc_logprob_vec.reserve(micro_batches_num);
+      for (auto i = 0; i < micro_batches_num; ++i) {
+        acc_logprob_vec.push_back(
+            batched_fwd_inputs.micro_inputs[i].acc_logprob);
+      }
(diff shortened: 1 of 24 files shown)


xllm/core/distributed_runtime/worker_service.cpp
```diff
@@ -346,6 +346,16 @@ void WorkerService::ExecuteModel(
...
 
+    // concat acc_logprob here for beam search together
+    batched_fwd_inputs.acc_logprob =
+        batched_fwd_inputs.micro_inputs[0].acc_logprob;
+    for (auto i = 1; i < micro_batches_num; ++i) {
+      batched_fwd_inputs.acc_logprob =
+          torch::cat({batched_fwd_inputs.acc_logprob,
+                      batched_fwd_inputs.micro_inputs[i].acc_logprob},
+                     -1);
+    }
```
RobbieLeung, line 361: Consider using `std::vector<torch::Tensor> acc_logprob_vec;` and `torch::stack(acc_logprob_vec);` for better performance.

xllm/core/runtime/worker_impl.h
```diff
@@ -230,6 +231,8 @@ class WorkerImpl {
 
   std::unique_ptr<Sampler> sampler_;
 
+  std::unique_ptr<BeamSearcher> beam_searcher_;
```
RobbieLeung: Move it to `LLMWorkerImpl` since beam searcher is only worked for llm.

xllm/core/distributed_runtime/worker_service.cpp
yinjiawei01: Updated.

RobbieLeung (approved):
LGTM

PR #285, opened 2025-10-27
feat: support mlu dp and ep and fix rope_forward.

a120092009:
1. fix rope forward
2. support mlu dp and ep

### DIFF
--- a/xllm/core/layers/common/qwen3_decoder_layer.cpp
+++ b/xllm/core/layers/common/qwen3_decoder_layer.cpp
@@ -69,8 +69,8 @@ torch::Tensor Qwen3DecoderImpl::forward(torch::Tensor& x,
                                         const AttentionMetadata& attn_metadata,
                                         KVCache& kv_cache,
                                         const ModelInputParams& input_params) {
-  bool is_dummy_run = layer::is_dummy_run(input_params, parallel_args_);
-  if (is_dummy_run) {
+  bool is_dummy = is_dummy_run(input_params, parallel_args_);
+  if (is_dummy) {
     return x;
   }
   // Pre-attention norm


xllm/core/framework/sampling/logits_utils.h
```diff
@@ -21,6 +21,8 @@ limitations under the License.
 
 #if defined(USE_NPU)
 #include "kernels/npu/xllm_ops/top_k_top_p.h"
+#elif defined(USE_MLU)
+#include "kernels/ops_api.h"
```
XuZhang99: no need to place this inside `#if defined`

xllm/core/layers/common/layer_utils.h
```diff
@@ -0,0 +1,31 @@
...
+==============================================================================*/
+
+#pragma once
+#include "framework/model/model_input_params.h"
+#include "framework/parallel_state/parallel_args.h"
+
+namespace xllm {
+namespace layer {
+
+bool is_dummy_run(const ModelInputParams& input_params,
```
XuZhang99, line 23: this dummy run only used for moe? if so, it's better to rename `is_moe_dummy_run`.

xllm/core/layers/common/qwen3_decoder_layer.cpp
```diff
@@ -66,6 +69,10 @@ torch::Tensor Qwen3DecoderImpl::forward(torch::Tensor& x,
                                         const AttentionMetadata& attn_metadata,
                                         KVCache& kv_cache,
                                         const ModelInputParams& input_params) {
+  bool is_dummy_run = layer::is_dummy_run(input_params, parallel_args_);
```
XuZhang99: `layer::` is useless, beacause they are in the same namespace.

xllm/core/layers/common/fused_moe.cpp
```diff
@@ -124,17 +120,41 @@ torch::Tensor FusedMoEImpl::forward_expert(
...
+  if (parallel_args_.ep_size() > 1) {
+    final_hidden_states = parallel_state::reduce(final_hidden_states,
+                                                 parallel_args_.moe_ep_group_);
+  }
   return final_hidden_states;
 }
 
-torch::Tensor FusedMoEImpl::forward(const torch::Tensor& hidden_states) {
+torch::Tensor FusedMoEImpl::forward(const torch::Tensor& hidden_states,
+                                    const ModelInputParams& input_params) {
```
XuZhang99, line 131: if we only use `input_params.dp_global_token_nums`, even in the future, how about replace `input_params` with `dp_global_token_nums` ?

PR #315, opened 2025-11-03
bugfix: resolve stream output parser core when dont enable function call or reasoning output.

RobbieLeung:


yingxudeng:
LGTM

yq33victor (approved):
LGTM

PR #316, opened 2025-11-04
feat: add mix instance role for SLO aware scheduling policy.

JimHsiung:


xllm/core/scheduler/disagg_pd_scheduler.cpp
```diff
@@ -125,16 +127,53 @@ void DisaggPDScheduler::profile_ttft() {
...
+       token_length >>= 1) {
+    max_batch_size = num_blocks / (token_length / block_size + 1);
+    int32_t current_max_batch_size = max_batch_size > max_seqs_per_batch
+                                         ? max_seqs_per_batch
+                                         : max_batch_size;
+    for (int32_t batch_size = current_max_batch_size; batch_size > 0;
+         batch_size *= 0.9) {
+      double latency = profile_manager_->profile_decode_step_time(
+          token_length, batch_size, /*min_context_len=*/64, max_context_len);
+      instance_info_.tpot_profiling_data.emplace_back(
```
weizhehuang0827, line 171: Can we directly send the locally trained parameters? Only about three values

JimHsiung: This is for generating random requests with lengths ranging between [min_context_len, max_context_len] to profile step execution time, so these two additional parameters need to be passed.

weizhehuang0827: I mean , can the coefficients of the linear model be directly transmitted in the data uploaded to `xllm_service` (in `instance_info`)?

xllm/core/runtime/llm_worker_impl.cpp
```diff
@@ -84,8 +84,7 @@ std::optional<ForwardOutput> LLMWorkerImpl::step(
     input_params_micro_batches.push_back(
         std::move(inputs.micro_inputs[i].input_params));
 
-    if (options_.instance_role() == InstanceRole::PREFILL &&
-        options_.kv_cache_transfer_mode() == "PUSH" &&
+    if (options_.kv_cache_transfer_mode() == "PUSH" &&
```
yq33victor, line 87: will this code also be executed in decode instance ?

JimHsiung: No, because the requests in the decode instances do not contain transfer_kv_infos.

xllm/core/scheduler/disagg_pd_scheduler.cpp
JimHsiung: Of course, that’s a good idea. However, the current scheduling strategy is not fully fixed, so keeping the original data is beneficial for potential changes to the strategy. If it is confirmed later that only coefficients are needed, we can modify this part of the logic.

xllm/core/scheduler/disagg_pd_scheduler.cpp
```diff
@@ -114,6 +115,7 @@ void DisaggPDScheduler::register_instance_info(const std::string& server_name,
 }
 
 void DisaggPDScheduler::profile_ttft() {
+  LOG(INFO) << "Start profiling TTFT.";
```
RobbieLeung, line 118: Remove debug info.

xllm/core/scheduler/profile/profile_manager.cpp
```diff
@@ -604,4 +604,69 @@ double ProfileManager::run_request(
...
+
+    std::uniform_int_distribution<int> dis(0, max);
+    int add = dis(gen);
+    token_length_vec[i] += add;
+    prefix_length_vec[i] += add;
+    remain -= add;
+  }
+
+  int idx = 0;
+  while (remain > 0) {
```
RobbieLeung, line 663: A risk of an infinite loop here.

xllm/core/scheduler/disagg_pd_scheduler.cpp
```diff
@@ -700,12 +745,32 @@ bool DisaggPDScheduler::decode_recv_first_generation(
                             dst_block_ids);
   }
 
+  // update latest_generate_time_ for sequence
+  request->sequences()[0]->tbt(absl::Now());
```
RobbieLeung, line 749: only need to process the first sequence? How about the others?

xllm/core/scheduler/disagg_pd_scheduler.cpp
JimHsiung: Since the profiling phase takes a long time, this log is added to indicate that the service is still in the startup process.

xllm/core/scheduler/disagg_pd_scheduler.cpp
JimHsiung: Disaggregated PD currently supports only one sequence.

xllm/core/scheduler/profile/profile_manager.cpp
JimHsiung: A check has been added to prevent infinite loops.

RobbieLeung (approved):

PR #318, opened 2025-11-05
feat: support embedding offline inference.

yq33victor:


examples/generate_embedding.py
```diff
@@ -0,0 +1,32 @@
...
+
+# Create an EmbeddingLM.
+parser = ArgumentParser()
+emb = Embedding(**vars(parser.parse_args()))
+
+# Create a reqeust params, include sampling params
+request_params = RequestParams()
+request_params.temperature = 0.8
+request_params.top_p = 0.95
+request_params.max_tokens = 10
```
JimHsiung: Do embedding models also need these request parameters?

RobbieLeung: > Do embedding models also need these request parameters?

Embedding models dont have sample process，so these request parameters are useless.

RobbieLeung: And `max_tokens` need to be 1, since embeding model only forward once.

yq33victor: ok , just refine the example!

Issue #320, opened 2025-11-05
这个代码怎么编的过的


laneeeee:
<img width="832" height="321" alt="Image" src="https://github.com/user-attachments/assets/274e99be-ae92-4afa-a8ba-238e7b6db6fd" />

<img width="763" height="270" alt="Image" src="https://github.com/user-attachments/assets/dbc924d0-61bd-4a5e-ac44-e0ccb32e3d74" />

<img width="692" height="309" alt="Image" src="https://github.com/user-attachments/assets/5f8fbb96-2a47-469c-9fa2-acb605767f3a" />

torch::Tensor forward定义和实现都不一样

yq33victor:
the function is not used, we will delete it.

laneeeee:
> the function is not used, we will delete it.

感谢你的解答，我使用g++编译的时候没有这个问题，但是在使用clang++的时候遇到了

PR #323, opened 2025-11-06
bugfix: fix the compilation error caused by override a base class function that does not exist.

yq33victor:


yq33victor:
fix #320

RobbieLeung (approved):

PR #331, opened 2025-11-07
feat: add timeout env variable for process groups.

a120092009:


phantomlei3:
**feature discription:**
add XLLM_PROCESS_GROUP_ASYNC_TIMEOUT_SECONDS environment variable (default: 4seconds) in llm_engine, dit_engine, and vlm_engine. This is particularly important in multi-node multi-device scenarios where network latency may require a longer timeout period.

liutongxuan (approved):

PR #334, opened 2025-11-07
bugfix: temporarily use heap allocation for VLM backend.

wly-115:
bugfix: temporarily use heap allocation for VLM backend.

VLM backend has memory growth issue with arena allocation.
Temporarily use new instead of arena as a workaround.


yq33victor (approved):
LGTM

RobbieLeung (approved):

PR #344, opened 2025-11-10
bugfix: eliminate use-after-move undefined behavior.

yingxudeng:


xllm/api_service/chat_service_impl.cpp
```diff
@@ -633,12 +637,12 @@ void ChatServiceImpl::process_async_impl(std::shared_ptr<ChatCall> call) {
       [call,
        model,
        master = master_,
-       stream = request_params.streaming,
+       stream = std::move(saved_streaming),
```
walsonyang, line 640: A simple bool variable, not necessary to use std::move

PR #374, opened 2025-11-13
feat: add cos_sin cache manager for rope to reduce memory footprint.

a120092009:
same logic with vllm：https://github.com/vllm-project/vllm/blob/releases/v0.9.1/vllm/model_executor/layers/rotary_embedding.py#L1702

cac6fc3ed4df1ece66bb0f4ebfc09eb7

0afe5407ad6f55916c7f8d1ad28de4aa

### DIFF
--- a/xllm/core/layers/rotary_embedding.cpp
+++ b/xllm/core/layers/rotary_embedding.cpp
@@ -28,6 +28,53 @@ limitations under the License.
 namespace xllm {
 
 namespace {
+
+// Hash function for CosSinCacheDesc
+struct CosSinCacheDescHash {
+  size_t operator()(const rotary::CosSinCacheDesc& key) const {
+    size_t hash = 0;
+    hash = rotary::hash_combine(hash, std::hash<int64_t>{}(key.rotary_dim));
+    hash = rotary::hash_combine(
+        hash, std::hash<int64_t>{}(key.max_position_embeddings));
+    hash = rotary::hash_combine(hash, std::hash<bool>{}(key.interleaved));
+    hash = rotary::hash_combine(hash, std::hash<float>{}(key.scaling_factor));
+    hash = rotary::hash_combine(hash, std::hash<float>{}(key.attn_factor));
+    hash = rotary::hash_combine(hash, std::hash<float>{}(key.mscale));
+    hash = rotary::hash_combine(hash, std::hash<float>{}(key.mscale_all_dim));
+    hash = rotary::hash_combine(hash, key.inv_freq_hash);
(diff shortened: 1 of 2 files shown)


xllm/core/layers/rotary_embedding.cpp
```diff
@@ -165,26 +232,42 @@ torch::Tensor compute_cos_sin_cache(int64_t rotary_dim,
...
+
+  // Get or create cache using cache manager
+  auto& cache_manager = CosSinCacheManager::get_instance();
+  return cache_manager.get_or_create_cache(key, [&]() {
+    float mscale_ = static_cast<float>(
+        yarn_get_mscale(scaling_factor, mscale) /
+        yarn_get_mscale(scaling_factor, mscale_all_dim) * attn_factor);
+    // [max_position_embeddings]
+    auto t = torch::arange(max_position_embeddings * scaling_factor);
+    // [max_position_embeddings, rotary_dim/2]
```
XuZhang99: These codes (from line 249 to 269) should be placed into a separate function.

xllm/core/layers/rotary_embedding.h
```diff
@@ -53,6 +55,40 @@ torch::Tensor compute_cos_sin_cache(int64_t rotary_dim,
                                     bool interleaved,
                                     torch::Tensor inv_freq,
                                     const torch::TensorOptions& options);
+
+// Internal: Cache key structure for cos_sin_cache sharing
+struct CosSinCacheKey {
```
XuZhang99: why call it `Key`? I think `Desc` is better.

xllm/core/layers/rotary_embedding.cpp
```diff
@@ -20,14 +20,62 @@ limitations under the License.
 
 #include <boost/algorithm/string.hpp>
 #include <cmath>
+#include <functional>
 #include <memory>
+#include <unordered_map>
```
XuZhang99: no need to 	include `functional` and `unordered_map` because they are already included in .h file.

xllm/core/layers/rotary_embedding.cpp
phantomlei3: fixed

xllm/core/layers/rotary_embedding.h
phantomlei3: renamed

xllm/core/layers/rotary_embedding.cpp
phantomlei3: fixed

XuZhang99 (approved):

yq33victor (approved):

PR #400, opened 2025-11-18
add WAN2.1-I2V model modules

haotianxu216:
The umt5_encoder.h and clip_vision_model.h modules have been registered as VLM models and tested, but issues were encountered during testing:
umt5_encoder.h error: engine.h:47 kv_cache_manager_ is not BlockManagerPool type!
clip_vision_model.h error: utils.cpp:32 Check failed: weight.sizes() == tensor.sizes() ([0, 1280] vs. [1280, 1280]) weight size mismatch for vision_model.encoder.layers.0.self_attn.q_proj.weight
The other three modules (autoencoder_kl_wan.h, dit_wan.h, and unipc_multistep_scheduler.h) have not been tested yet. Additionally, the overall pipeline for the WAN model has not been implemented.

### DIFF
--- a/xllm/models/dit/clip_vision_model.h
+++ b/xllm/models/dit/clip_vision_model.h
@@ -1,0 +1,266 @@
+#pragma once
+
+#include <atb/atb_infer.h>
+#include <c10/core/ScalarType.h>
+#include <torch/torch.h>
+
+#include <regex>
+#include <unordered_map>
+
+#include "clip_text_model.h"
+#include "core/framework/dit_model_loader.h"
+#include "core/framework/kv_cache/kv_cache.h"
+#include "core/framework/model/model_input_params.h"
+#include "core/framework/model_context.h"
+#include "core/layers/siglip_encoder_layer.h"
+#include "dit_linear.h"
+#include "models/model_registry.h"
+#include "processors/clip_image_processor.h"
+#include "processors/input_processor.h"
+#include "processors/pywarpper_image_processor.h"
+#include "xllm_kernels/core/include/atb_speed/log.h"
+
+namespace xllm {
+
+class CLIPVisionEmbeddingImpl : public torch::nn::Module {
+ public:
+  explicit CLIPVisionEmbeddingImpl(const ModelContext& context) {
+    auto model_args = context.get_model_args();
+    auto options = context.get_tensor_options();
+    embed_dim_ = model_args.mm_hidden_size();
+    image_size_ = model_args.mm_image_size();
+    class_embedding_ = register_parameter("class_embedding",
+                                          torch::randn({embed_dim_}, options));
+    patch_embedding_ = register_module(
+        "patch_embedding",
+        torch::nn::Conv2d(torch::nn::Conv2dOptions(model_args.mm_num_channels(),
+                                                   embed_dim_,
+                                                   model_args.mm_patch_size())
+                              .stride(model_args.mm_patch_size())
+                              .bias(false)));
+    patch_embedding_->weight.set_data(patch_embedding_->weight.to(options));
+
+    auto num_patches =
+        (model_args.mm_image_size() / model_args.mm_patch_size()) *
+        (model_args.mm_image_size() / model_args.mm_patch_size());
+    auto num_positions = num_patches + 1;
+    position_embedding_ =
+        register_parameter("position_embedding",
+                           torch::randn({num_positions, embed_dim_}, options));
+    position_ids_ = register_buffer(
+        "position_ids",
+        torch::arange(0, num_positions, torch::kLong).unsqueeze(0));
+  }
+
+  torch::Tensor forward(const torch::Tensor& pixel_values) {
+    int64_t batch_size = pixel_values.size(0);
+    auto patch_embeds =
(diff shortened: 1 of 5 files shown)


xllm/models/dit/clip_vision_model.h
```diff
@@ -0,0 +1,266 @@
...
+#include "processors/input_processor.h"
+#include "processors/pywarpper_image_processor.h"
+#include "xllm_kernels/core/include/atb_speed/log.h"
+
+namespace xllm {
+
+class CLIPVisionEmbeddingImpl : public torch::nn::Module {
+ public:
+  explicit CLIPVisionEmbeddingImpl(const ModelContext& context) {
+    auto model_args = context.get_model_args();
```
liujinguang0125, line 28: why not use "const auto& model_args = xxx" implementation method here, it has good performance because it doesn't copy the return object. options variable has the same issue.

xllm/models/dit/clip_vision_model.h
```diff
@@ -0,0 +1,266 @@
...
+    auto options = context.get_tensor_options();
+    embed_dim_ = model_args.mm_hidden_size();
+    image_size_ = model_args.mm_image_size();
+    class_embedding_ = register_parameter("class_embedding",
```
liujinguang0125, line 32: I think it's meaningless to use random initialization for class_embedding_ and position_embedding_, they will be overwritten by load_state_dict function, it's recommended to use torch::zeros.

xllm/models/dit/clip_vision_model.h
```diff
@@ -0,0 +1,266 @@
...
+                              .stride(model_args.mm_patch_size())
+                              .bias(false)));
+    patch_embedding_->weight.set_data(patch_embedding_->weight.to(options));
+
+    auto num_patches =
+        (model_args.mm_image_size() / model_args.mm_patch_size()) *
+        (model_args.mm_image_size() / model_args.mm_patch_size());
+    auto num_positions = num_patches + 1;
+    position_embedding_ =
+        register_parameter("position_embedding",
```
liujinguang0125, line 48: the function prototype of register_parameter is as following：
template <typename T>
Tensor& register_parameter(const std::string& name, T&& tensor, bool requires_grad = true);

the requires_grad parameter should be set to false for inferrence service and it will take less memory and can be optimized deeply. or may it use register_buffer here?

xllm/models/dit/clip_vision_model.h
```diff
@@ -0,0 +1,266 @@
...
+    transformer_ =
+        register_module("transformer", CLIPVisionTransformer(context));
+    projection_ = register_parameter(
+        "projection",
+        torch::randn(
+            {model_args.mm_hidden_size(), model_args.mm_projection_dim()},
+            options));
+  }
+
+  torch::Tensor forward(const torch::Tensor& input_ids) {
```
liujinguang0125, line 211: change the name "input_ids" to "pixel_values"?

xllm/models/dit/clip_vision_model.h
```diff
@@ -0,0 +1,266 @@
...
+    CHECK(post_layernorm_bias_loaded_)
+        << "weight is not loaded for " << prefix + "post_layernorm.bias";
+  }
+
+ private:
+  CLIPVisionEmbedding embeddings_{nullptr};
+  torch::nn::LayerNorm pre_layernorm_{nullptr};
+  CLIPEncoder encoder_{nullptr};
+  torch::nn::LayerNorm post_layernorm_{nullptr};
+  bool pre_layernorm_weight_loaded_ = false;
```
liujinguang0125, line 190: The naming of bool type variables is not uniform, add prefix "is_" just as these variables in CLIPVisionEmbeddingImpl.

xllm/models/dit/clip_vision_model.h
```diff
@@ -0,0 +1,266 @@
...
+
+  // load the weight from the checkpoint
+  void load_state_dict(const StateDict& state_dict) {
+    transformer_->load_state_dict(
+        state_dict.get_dict_with_prefix("vision_model."));
+    const auto proj_weight = state_dict.get_tensor("projection");
+    if (proj_weight.defined()) {
+      DCHECK_EQ(projection_.sizes(), proj_weight.sizes())
+          << "projection size mismatch for " << name();
+      projection_.data().copy_(proj_weight);
```
liujinguang0125, line 225: projection_ will use random weights if proj_weight is invalid, it's better to add booltype varaible and check it.

xllm/models/dit/clip_vision_model.h
```diff
@@ -0,0 +1,266 @@
...
+    auto embeddings = torch::cat({class_embeds, patch_embeds}, 1);
+    embeddings += position_embedding_.index({position_ids_});
+    return embeddings;
+  }
+
+  // load the weight from the checkpoint
+  void load_state_dict(const StateDict& state_dict) {
+    const auto cls = state_dict.get_tensor("class_embedding");
+    if (cls.defined()) {
+      DCHECK_EQ(cls.sizes(), class_embedding_.sizes())
```
liujinguang0125, line 69: CHECK_EQ will be better than DCHECK_EQ which takes effect only in debug mode.

PR #424, opened 2025-11-24
refactor: move validation of mla from engine to model layer.

Hermit-w:


xllm/core/layers/npu/npu_deepseek_v2_decoder_layer_impl.cpp
```diff
@@ -1540,7 +1540,12 @@ torch::Tensor NpuDeepseekV2DecoderLayerImpl::forward(
     LOG_IF(FATAL, st != 0) << model_name_
                            << "excute prefill layer fail, error code: " << st;
   } else {
-    if (!FLAGS_enable_customize_mla_kernel) {
+    const int num_tokens = x.sizes().at(0);
+    // decode phase with tokens more than this limit will lead to error in
+    // customize mla kernel. once detect any input exceed the limit, fall back
+    // to default kernel.
+    const int num_tokens_limit = 230;
+    if (!FLAGS_enable_customize_mla_kernel || num_tokens >= num_tokens_limit) {
```
LMX-xin: decode_mla_node is the same as decode_node but with an additional enableCustomizeMla parameter. So, can decode_node be reused without creating a new decode_mla_node?

PR #435, opened 2025-11-25
feat: implement chunked prefill and prefix cache for Qwen3 MoE.

yingxudeng:


yingxudeng:
TODO：
We also need to package the new version of xllm_kernels after the machine has recovered, because the ATB group diagram has undergone some changes.

yingxudeng:
The xllm_kernels package has been updated.

RobbieLeung (approved):

PR #443, opened 2025-11-26
feat: update Docker image to support ATB, ACLNN fused ops and ProcessGroupHCCL.

yingxudeng:


yingxudeng:
I have updated the base image on the main branch. This merge is essential; otherwise, the build will fail with the new image. Could you please review and merge this code change?    @RobbieLeung @liutongxuan

PR #445, opened 2025-11-27
Bugfix kvcache store

Kang-Meng:


xllm/core/framework/request/sequence.h
```diff
@@ -74,6 +75,44 @@ struct SequenceParams {
...
+
+      return elapsed.count() >= timeout_ms;
+    }
+  }
+
+  void reset() { is_timeout_set_ = false; }
+
+ private:
+  static void init_timeout() {
+    const char* env_str = std::getenv("PREFETCH_TIMEOUT_MS");
```
liujinguang0125, line 106: there are two issues about TimeoutChecker implementation:
1. does a gflag variable work here, why it introduces a env variable? 
2. timeout_ms should be an static attribute of the class, not a global variable.

xllm/core/framework/request/sequence.h
```diff
@@ -74,6 +75,44 @@ struct SequenceParams {
...
 };
 
+static uint32_t timeout_ms = 0;
+class TimeoutChecker {
+ private:
+  std::chrono::steady_clock::time_point timeout_start_;
+  bool is_timeout_set_ = false;
+
+ public:
+  TimeoutChecker() { init(); }
```
liujinguang0125, line 85: can this implementation improve easy of use as following:

  TimeoutChecker() { 
    timeout_start_ = std::chrono::steady_clock::now();
   }

  bool check_timeout() {
      auto now = std::chrono::steady_clock::now();
      auto elapsed = std::chrono::duration_cast<std::chrono::milliseconds>(
          now - timeout_start_);

      return elapsed.count() >= timeout_ms;
  }

  void reset() { timeout_start_ = std::chrono::steady_clock::now(); }

PR #449, opened 2025-11-27
refactor: optimize unique token count preparation of batch input builder.

RobbieLeung:
The `unique_token_counts` parameter has been updated.  
Note that:  
- `presence_penalty` and `frequency_penalty` only apply to **content tokens** and their **occurrence counts**.  
- `repetition_penalty`, however, affects **both prompt and content tokens** but does **not** consider their counts.


yq33victor (dismissed):
LGTM

JimHsiung (approved):

PR #454, opened 2025-11-28
feat: support deepseek mtp on mlu.

a120092009:


xllm/core/runtime/speculative_worker_impl.cpp
```diff
@@ -36,6 +36,101 @@ namespace {
...
+                            int32_t offset) {
+#if defined(USE_NPU)
+  kv_seq_lens_vec.emplace_back(kv_seq_lens_slice[seq_id] + offset);
+#else
+  // build cumulative format for kv_seq_lens
+  int32_t offset_kv_len =
+      kv_seq_lens_slice[seq_id + 1] - kv_seq_lens_slice[seq_id] + offset;
+  push_cumsum(kv_seq_lens_vec, offset_kv_len);
+#endif
+}
```
RobbieLeung: The calculation of `kv_max_seq_len` should be moved here, and `get_kv_max_seq_len` should be removed.

phantomlei3: have been fixed

PR #456, opened 2025-11-29
feat: add NPU process group initialization and management.

yingxudeng:


yingxudeng:
image
We are currently encountering compilation issues when introducing processGroupHccl. We might need to update the image again.

xllm/core/common/CMakeLists.txt
```diff
@@ -29,6 +29,7 @@ cc_library(
     absl::random_random
     absl::strings
     torch
+    $<$<BOOL:${USE_NPU}>:torch_python>
```
XuZhang99: no need to add `USE_NPU`.

xllm/core/framework/parallel_state/npu_process_group.cpp
```diff
@@ -24,113 +34,65 @@ namespace {
...
+                                   int rank_size,
+                                   int port,
+                                   bool trans,
+                                   const std::string& host,
+                                   const std::string& group_name,
+                                   const torch::Device& device)
+    : ProcessGroup(device) {
+  c10::intrusive_ptr<c10d_npu::ProcessGroupHCCL::Options> hccl_pg_options =
+      c10d_npu::ProcessGroupHCCL::Options::create();
+  // hccl_pg_options->group_name = group_name;
```
XuZhang99:
```cpp
#if TORCH_VERSION_MAJOR >= 2 && TORCH_VERSION_MINOR >= 7
    pg_options->group_name = group_name;
#endif 
```

xllm/core/framework/parallel_state/collective_communicator.cpp
```diff
@@ -18,6 +18,7 @@ limitations under the License.
 #include "mapping_npu.h"
 
 #if defined(USE_NPU)
+#include "npu_process_group.h"
 #include "xllm_kernels/core/include/atb_speed/base/external_comm_manager.h"
 #include "xllm_kernels/core/include/atb_speed/utils/singleton.h"
 #include "xllm_kernels/models/base/param/mapping.h"
```
XuZhang99: line 22-24 seems useless?

xllm/core/framework/parallel_state/npu_process_group.cpp
```diff
@@ -14,6 +14,16 @@ limitations under the License.
 ==============================================================================*/
 
 #include "npu_process_group.h"
```
XuZhang99, line 16: npu_process_group.cpp should be deleted, because npu_process_group.h is enough, like cuda/mlu_process_group.h.

xllm/core/framework/parallel_state/npu_process_group.cpp
```diff
@@ -24,113 +34,65 @@ namespace {
...
-  //     /*recvbuff=*/flattened_output.data_ptr(),
-  //     /*sendcount=*/count,
-  //     /*datatype=*/data_type,
-  //     /*comm=*/comm_,
-  //     /*stream=*/stream));
-  // // copy the flattened output tensors to the outputs.
-  // for (int i = 0; i < outputs.size(); ++i) {
-  //   outputs[i].copy_(flattened_output[i], /*non_blocking=*/true);
-  // }
+std::unique_ptr<xllm::ProcessGroup> create_process_group(
```
XuZhang99: `create_process_group` function can placed into an anonymous namespace in collective_communicator.cpp for all devices.

xllm/core/framework/parallel_state/npu_process_group.cpp
```diff
@@ -14,6 +14,16 @@ limitations under the License.
 ==============================================================================*/
 
 #include "npu_process_group.h"
+#ifdef TORCH_HIGHER_THAN_PTA6
+#include <torch_npu/csrc/framework/OpCommand.h>
+#else
+#include <torch_npu/csrc/aten/NPUNativeFunctions.h>
+#include <torch_npu/csrc/framework/utils/OpPreparation.h>
+#endif
```
XuZhang99: Do we really need these headers from line 17 to line 22?

xllm/core/framework/parallel_state/npu_process_group.cpp
yingxudeng: Thanks for the suggestion! I'll add the version check.

To ensure forward compatibility (e.g., for PyTorch 3.0 where MINOR might be 0), I'll adjust the logic to cover cases where MAJOR > 2 as well

xllm/core/framework/parallel_state/collective_communicator.cpp
yingxudeng: Thanks for the review. These lines were not added by me, but I verified that lines 22-23 are actually necessary for atb_speed::GetSingleton. However, line 24 is indeed unused, so I will remove it.

xllm/core/framework/parallel_state/npu_process_group.cpp
yingxudeng: Thanks for the review. However, I strongly prefer to keep the .cpp file. Defining implementation directly in the header is generally considered bad practice. I hope you understand my decision to maintain this separation.

xllm/core/framework/parallel_state/npu_process_group.cpp
yingxudeng: Thanks for the feedback.

Regarding the suggestion to consolidate the create_process_group functions, I have a few concerns:

Since ProcessGroupHCCL, ProcessGroupCncl, and ProcessGroupNccl are device-specific implementations , moving them to collective_communicator.cpp would introduce excessive #if/#elif preprocessor directives.

xllm/core/common/global_flags.cpp
```diff
@@ -430,4 +430,9 @@ DEFINE_bool(
     enable_dp_balance,
     false,
     "Whether to enable dp load balance, if true, sequences within a single "
-    "dp batch will be shuffled.");
\ No newline at end of file
+    "dp batch will be shuffled.");
+
+DEFINE_string(
```
liujinguang0125, line 469: does it need to use USE_NPU macro here, since it's only used in NPU platform.

xllm/core/common/CMakeLists.txt
yingxudeng: Thanks for the review. I've addressed your comments.

xllm/core/framework/parallel_state/npu_process_group.cpp
yingxudeng: Thanks for the review. I've addressed your comments.

xllm/core/common/global_flags.cpp
yingxudeng: Thanks for the review. I've addressed your comments.

yingxudeng:
I have updated the three images on the main branch, and the build is now passing. Would you mind taking a look at the code to see if it is ready for merging? I would appreciate your review.   @yq33victor

xllm/core/framework/parallel_state/npu_process_group.cpp
```diff
@@ -24,113 +28,68 @@ namespace {
...
-  //     /*recvbuff=*/flattened_output.data_ptr(),
-  //     /*sendcount=*/count,
-  //     /*datatype=*/data_type,
-  //     /*comm=*/comm_,
-  //     /*stream=*/stream));
-  // // copy the flattened output tensors to the outputs.
-  // for (int i = 0; i < outputs.size(); ++i) {
-  //   outputs[i].copy_(flattened_output[i], /*non_blocking=*/true);
-  // }
+std::unique_ptr<xllm::ProcessGroup> create_process_group(
```
yq33victor: nit: here we use a template to solve it:
```
template<ProcessGroupType>
std::unique_ptr<xllm::ProcessGroup> create_process_group(
    int32_t rank,
    int32_t world_size,
    int32_t rank_size,
    int32_t port,
    bool trans,
    const std::string& host,
    const std::string& group_name,
    const torch::Device& device) {
  return std::make_unique<ProcessGroupType>(
      rank, world_size, rank_size, port, trans, host, group_name, device);
}
```

xllm/core/framework/parallel_state/cuda_process_group.h
```diff
@@ -21,23 +21,24 @@ limitations under the License.
...
+                   int32_t port,
                    bool trans,
                    const std::string& host,
                    const std::string& group_name,
                    const torch::Device& device)
       : ProcessGroup(device) {
     c10::intrusive_ptr<c10d::ProcessGroupNCCL::Options> pg_options =
         c10d::ProcessGroupNCCL::Options::create();
-#if TORCH_VERSION_MAJOR >= 2 && TORCH_VERSION_MINOR >= 7
+#if TORCH_VERSION_MAJOR > 2 || \
```
XuZhang99, line 37: the latest version of pytorch is 2.9.0, so no need to add `TORCH_VERSION_MAJOR > 2`.

yingxudeng: While the current logic works for now, we should consider potential future releases, such as Torch 3.x. Adding this check would make the implementation more robust and future-proof.

PR #475, opened 2025-12-03
feat: add c_api interface for llm inference.

magicheng0816:


xllm/cc_api/types.h
```diff
@@ -0,0 +1,320 @@
...
+  // the unix timestamp (in seconds) of when the completion was created.
+  int64_t created;
+
+  // the model used for the completion
+  std::string model;
+
+  // list of generated completion choices for the input prompt
+  std::vector<XLLM_Choice> choices;
+
+  // usage statistics for the completion request.
```
liujinguang0125: The first letter of each sentence should be capitalized. Each sentence must end with a period.

xllm/cc_api/types.h
```diff
@@ -0,0 +1,320 @@
...
+
+  // draft hf model path to the model file
+  std::optional<std::string> draft_model = std::nullopt;
+
+  // Devices to run the draft model on, e.g. npu:0, npu:0,npu:1
+  std::optional<std::string> draft_devices = std::nullopt;
+};
+
+struct XLLM_CAPI_EXPORT XLLM_RequestParams {
+  // whether to include the original prompt in the response. default = true
```
liujinguang0125: if the default value is true, the echo attribute needs to be initialized.

PR #476, opened 2025-12-03
refactor: breakdown fused moe kernel for deepseek all2all setup.

a120092009:


xllm/core/layers/common/indexer.h
```diff
@@ -25,7 +25,8 @@ limitations under the License.
 #include "../mlu/attention.h"
 #elif defined(USE_CUDA)
 #include "../cuda/attention.h"
-#endif #include "framework/kv_cache/kv_cache.h"
+#endif
+#include "framework/kv_cache/kv_cache.h"
```
XuZhang99: line 29 can be deleted.

phantomlei3: fixed

PR #480, opened 2025-12-03
feat: add constrained decoding for generative recommendation

magicheng0816:


xllm/core/framework/sampling/constrained_decoding.h
```diff
@@ -0,0 +1,36 @@
...
+// conforms to specific formats or rules.
+class ConstrainedDecoding {
+ public:
+  virtual ~ConstrainedDecoding();
+
+  virtual bool build_mask_cache();
+
+  // Input generated_token_list: [sequence_num][generated_token_ids]
+  // Output: mask tensor[sequence_num,vocab_size]
+  virtual torch::Tensor generate_mask(
```
linkerzhang, line 44: There'll be a data copy when calling this function, right? if it's heavy, it should be avoided.

xllm/core/framework/sampling/constrained_decoding.h
```diff
@@ -0,0 +1,36 @@
...
+
+namespace xllm {
+
+// Constrained decoding is used to ensure that the generated content
+// conforms to specific formats or rules.
+class ConstrainedDecoding {
+ public:
+  virtual ~ConstrainedDecoding();
+
+  virtual bool build_mask_cache();
```
linkerzhang: add comments please.

magicheng0816: done

xllm/core/framework/sampling/constrained_decoding.h
magicheng0816: There is no heavy data copying. First, an initialized tensor is generated on the device side based on the vocab size. Then, a valid token index set is dynamically obtained from the already generated tokens on the host side, copied to the device side, and the initialized tensor is modified in-place to form a mask.

xllm/core/framework/sampling/rec_constrained_decoding.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+constexpr float PRE_MASK_FACTOR = -10000.0f;
+constexpr int GEN_MASK_THREAD_NUM = 16;
+
+RecConstrainedDecoding::RecConstrainedDecoding(uint64_t model_version,
+                                               const int32_t vocab_size,
+                                               torch::ScalarType dtype,
+                                               torch::Device device,
+                                               bool use_gen_threadpool)
+    : model_version_(model_version),
```
liujinguang0125: There is a compiler warning because the order of the member initializer list is incorrect.

xllm/core/framework/sampling/rec_constrained_decoding.h
```diff
@@ -0,0 +1,54 @@
...
+
+  torch::Tensor generate_mask(
+      const std::vector<std::vector<int32_t>>& generated_token_list) override;
+
+ private:
+  torch::Tensor generate_decode_mask(
+      const std::vector<std::vector<int32_t>>& generated_token_list);
+
+ private:
+  bool build_mask_cache_;
```
liujinguang0125, line 48: it's better to add prefix "is_" for bool variable, such as is_build_mask_cache_

PR #490, opened 2025-12-05
bugfix: fix the issue of ineffective input embedding transmission.

magicheng0816:


xllm/core/framework/batch/batch_input_builder.cpp
```diff
@@ -688,6 +688,11 @@ RawForwardInput BatchInputBuilder::state_to_raw_forward_input() {
   process_swap_block_infos(raw_forward_input);
   raw_forward_input.batch_id = batch_id_;
 
+  if (input_embeddings_vec_.size() > 0) {
+    torch::Tensor input_embeddings = torch::cat(input_embeddings_vec_);
+    raw_forward_input.embeddings = tensor_to_2d_float_vector(input_embeddings);
+  }
```
yq33victor, line 694: nit: emmm, it seems that the `input_embedding` is useless now, maybe @wly-115 @yiming-l21 @RobbieLeung can help to confirm this. If so, we can delete all input_embedding related codes.

xllm/core/framework/batch/batch.cpp
```diff
@@ -48,7 +48,7 @@ void Batch::add(Sequence* sequence, uint32_t allowed_max_token) {
   allowed_max_tokens_.push_back(allowed_max_token);
 
   const auto& input_embedding = sequence->get_input_embedding();
-  if (input_embedding.defined())
+  if (sequence->stage() == SequenceStage::PREFILL && input_embedding.defined())
```
linkerzhang, line 51: will sequence be nullptr? if not, the signature should be "Sequence&",

linkerzhang: if yes, the nullptr case should be handled.

xllm/core/util/tensor_helper.h
```diff
@@ -50,6 +50,28 @@ inline torch::Tensor create_2d_tensor(const std::vector<std::vector<T> >& vec,
   return tensor;
 };
 
+inline std::vector<std::vector<float>> tensor_to_2d_float_vector(
```
linkerzhang, line 53: this will lead to a copy.... please fix it.

linkerzhang (changes requested):

xllm/core/framework/batch/batch_input_builder.cpp
magicheng0816: input_embedding will be used in generative recommendation, don't delete it for now

xllm/core/util/tensor_helper.h
magicheng0816: 1. When passing 'const torch::Tensor&tensor' as a function parameter, it does not trigger the copying of tensor data itself, only pass the 'reference' of tensor objects, and pytorch's torch:: Tensor is just an intelligent pointer container with reference counting. Passing torch:: Tensor tensor directly will also not result in data copying;
2. When 'std::vector<std::vector<float>>' is used as the return value of a function, there will be no data copying. Firstly, std::vector is a 'dynamic array' container, and calling the vector's move constructor only transfers ownership of the metadata. Secondly, the compiler will enable Named Return Value Optimization (NRVO), which directly constructs the vector at the memory address on the calling end, completely skipping copy/move;
3. Going back inside the function, if input tensor satisfies tensor.device().type()==torch:: kCPU&&tensor. scalar_type()== torch::kFloat32 && tensor.is_contiguous(), there will also be no data copying, but if one condition is not met, it will result in data copying, which is logical.

xllm/core/framework/batch/batch.cpp
magicheng0816: At the beginning of the function, ‘CHECK(sequence != nullptr)’, we will ensure that the sequence here will not be nullptr, and the use of pointers instead of references here is necessary to keep the sequence in both the batch and request.

PR #505, opened 2025-12-09
bugfix: support multiple models or multiple model version with independent instance.

liujinguang0125:
This pull request supports multiple models or multiple model version with independent instance for  cplusplus APIs.


xllm/core/distributed_runtime/worker_server.cpp
```diff
@@ -83,7 +83,7 @@ void WorkerServer::create_server(
 
   auto addr = net::get_local_ip_addr();
   auto worker_server =
-      ServerRegistry::get_instance().register_server("DistributeWorkerServer");
+      ServerRegistry::get_instance().register_server(server_name_);
```
magicheng0816, line 86: For singleton patterns like ServerRegistry, where multiple xllm instances use them, it feels more appropriate to construct a singleton pool bound to the xllm instance, rather than multiple xllms sharing the same ServerRegistry singleton.

xllm/core/distributed_runtime/worker_server.cpp
magicheng0816, line 297: If the user applies different XLLM instances to serve different models, then the 'global sharing' of gflags also needs to be modified to 'instance isolation'.

magicheng0816 (approved):

xllm/cc_api/examples/CMakeLists.txt
```diff
@@ -0,0 +1,52 @@
...
+target_include_directories(single_llm_instance
+    PRIVATE /usr/local/xllm/include
+)
+
+target_link_libraries(single_llm_instance
+    PRIVATE xllm
+)
+
+target_compile_definitions(single_llm_instance
+    PRIVATE _GLIBCXX_USE_CXX11_ABI=0
```
XuZhang99: the value of `_GLIBCXX_USE_CXX11_ABI` is different on different platform, it's up to the torch version. we can get its value through:
```python
   import torch
   print(torch.compiled_with_cxx11_abi())
   # or 
   print(torch._C._GLIBCXX_USE_CXX11_ABI)
```

xllm/core/distributed_runtime/worker_server.cpp
liujinguang0125: xLLM will support multiple models in future, then we can consider how to solve this kind of issue.

xllm/cc_api/examples/CMakeLists.txt
liujinguang0125: It has aligned with the related implementation of the main branch.

xllm/core/distributed_runtime/worker_server.cpp
liujinguang0125: as above comment, it needs to make big change to support this in current architecture. Now we only focus on the feature of supporting one mode with multiple versions in recommandation scene.

PR #507, opened 2025-12-09
feat: vlm support binary mm input.

xiao-yu-chen:


xllm/api_service/api_service.cpp
```diff
@@ -168,16 +186,17 @@ void ChatCompletionsImpl(std::unique_ptr<Service>& service,
   auto resp_pb =
       google::protobuf::Arena::CreateMessage<typename ChatCall::ResType>(arena);
 
-  std::string attachment = std::move(ctrl->request_attachment().to_string());
-  std::string error;
+  auto content_len = GetJsonContentLength(ctrl);
+  std::string attachment;
+  ctrl->request_attachment().copy_to(&attachment, content_len, 0);
```
yq33victor, line 188: what's the difference compare to the original code ?

xiao-yu-chen: binary can be appended after json. multimodal input and output can be optimized from base64 to binary, as well as tensor data。

yq33victor (approved):
LGTM

PR #519, opened 2025-12-11
feat: support fused norm for Qwen3 and DeepSeek.

XuZhang99:


xllm/core/layers/common/rms_norm.cpp
```diff
@@ -40,34 +40,47 @@ RMSNormImpl::RMSNormImpl(const ModelContext& context)
...
 
-torch::Tensor RMSNormImpl::forward_output(torch::Tensor& input,
-                                          torch::Tensor& output) {
-  auto org_shape = output.sizes().vec();
-  input = input.reshape({-1, norm_dim_});
-  output = output.reshape({-1, norm_dim_});
+  std::optional<torch::Tensor> residual_out;
+  if (residual.has_value()) {
+    residual.value() = residual.value().reshape({-1, norm_dim_});
+    if (Device::type_str() == "mlu") {
```
yq33victor, line 57: Consider `cuda` and other `cuda-like` device ?  so here we use:
```
if (Device::type_str() != "npu")  ?
```

XuZhang99: only mlu need `residual_out` for now.

yq33victor (approved):

PR #522, opened 2025-12-11
refactor: refactor runtime and distributed_runtime to break the circular dependency.

yq33victor:


yq33victor:
Currently, runtime and distributed_runtime have mutual dependencies, which in some cases lead to compilation issues—for example, the build problems we encountered yesterday when enabling DEBUG=1. To address this, I moved some code into distributed_runtime. The runtime module now retains only classes that are intended to operate on a single worker process, such as worker_impl, executor, etc.

@RobbieLeung @JimHsiung @DongheJin @XuZhang99  **need to review this PR**. cc @liutongxuan 

**NOTE:
We will not allow the runtime module to reference any code from distributed_runtime in the future.**

xllm/core/runtime/CMakeLists.txt
```diff
@@ -81,32 +81,32 @@ cc_library(
...
+#    :scheduler
+#    :request
+#    :runtime
+#    :model
+#    :models
+#    $<$<BOOL:${USE_NPU}>:npu_layers>
+#    :chat_template
+#    glog::glog
+#    $<$<BOOL:${USE_MLU}>:torch_mlu>
+#)
```
RobbieLeung: Remove unused code and dependencies.

XuZhang99 (dismissed):

RobbieLeung (dismissed):

JimHsiung (dismissed):

JimHsiung (approved):

PR #539, opened 2025-12-13
refactor: refactor mm_data module.

xiao-yu-chen:


xllm/core/framework/request/mm_batch_data.cpp
```diff
@@ -0,0 +1,121 @@
...
+    : ty_(ty), data_(std::move(items)) {}
+
+bool MMBatchData::has(const MMKey& key) const {
+  if (!valid()) return false;
+
+  const auto& itor = data_.find(key);
+  if (itor != data_.end())
+    return true;
+  else
+    return false;
```
XuZhang99: add `{}` for `if else`.

xllm/core/framework/request/mm_batch_data.cpp
```diff
@@ -0,0 +1,121 @@
...
+}
+
+void MMBatchData::get(const MMKey& key, std::vector<torch::Tensor>& vec) const {
+  if (!valid()) return;
+
+  const auto& itor = data_.find(key);
+  if (itor == data_.end()) return;
```
XuZhang99, line 41: `LOG(FATAL)` instead of `return;`.

xllm/core/util/tensor_helper.h
```diff
@@ -131,6 +131,68 @@ inline torch::Tensor safe_concat(const torch::Tensor& t1,
   }
 }
 
+inline bool safe_concat(const std::vector<torch::Tensor>& vec,
```
XuZhang99, line 134: too many `if` in this func, refactor it.

xllm/core/framework/request/mm_batch_data.cpp
xiao-yu-chen: for simplicity, modify it to “return itor = data_.find(key)；”

xllm/core/framework/request/mm_batch_data.cpp
xiao-yu-chen: it may be a bit strict, especially for online services. in addition, users of the class must complete initialization first

xllm/core/framework/request/mm_data.h
```diff
@@ -23,141 +23,160 @@ limitations under the License.
...
-struct VideoMetadata {
-  double fps = 0.0;              // original fps
-  int64_t total_num_frames = 0;  // original frames
-  double duration = 0.0;
-  double sampled_fps = 0.0;
-  torch::Tensor frame_indices;
-  std::vector<double> timestamps;
-};
+  bool has(uint32_t type) const { return type & ty_ != 0; }
+  bool has(MMType type) const { return type & ty_ != 0; }
```
DragonFive, line 66: Maybe (type & ty_) != 0 is better for read.

xllm/core/framework/request/mm_batch_data.h
```diff
@@ -0,0 +1,85 @@
...
+
+namespace xllm {
+
+class MMBatchData {
+ public:
+  MMBatchData() = default;
+  MMBatchData(const std::vector<MMData>& datas);
+  MMBatchData(uint32_t ty, const MMDict& items);
+
+  bool has(uint32_t type) const { return type & ty_ != 0; }
```
DragonFive, line 37: Maybe (type & ty_) != 0 is better for read.

xllm/core/framework/request/mm_data.h
```diff
@@ -23,141 +23,160 @@ limitations under the License.
...
+      }
+
+      torch::Tensor ts;
+      MMValue value;
+      if (safe_concat(lst, ts)) {
+        value = ts;
+      } else {
+        value = lst;
+      }
+      return std::get<T>(value);
```
DragonFive, line 131: I think may be there may be a potential bug, if I use  'get<torch::Tensor>(key)‘ , the function output type will be std::optional<torch::Tensor>, but if (safe_concat(lst, ts) return false, the value will be std::vector<torch::Tensor>, but use std::get<torch::Tensor>(std::vector<torch::Tensor>) will result in error ?

DragonFive: Suggestion: make get<T> robust by using std::get_if and
  returning nullopt on type mismatch, or explicitly discourage/disable templated get<T> for MMItemVec and
  require using get(key, vector<Tensor>&) + explicit caller-side handling/concat policy

xllm/core/framework/request/mm_data.h
xiao-yu-chen: the operator priority has been ensured to be correct, and the code is as concise as possible

xllm/core/framework/request/mm_batch_data.h
xiao-yu-chen: the operator priority has been ensured to be correct, and the code is as concise as possible

xllm/core/framework/request/mm_data.h
xiao-yu-chen: this is a good question，code hardening has been completed. in addition, the existing “void get(const MMKey& key, std::vector<torch::Tensor>& items) const” interface can be used to avoid ambiguity

PR #541, opened 2025-12-14
feat: set enable_mla by model type automatically.

RobbieLeung:


xllm/xllm.cpp
```diff
@@ -126,6 +139,19 @@ int run() {
     FLAGS_max_tokens_per_chunk_for_prefill = FLAGS_max_tokens_per_batch;
   }
 
+  // set enable_mla by model type
+  std::string model_type = get_model_type(model_path);
+  static std::unordered_set<std::string> deepseek_models = {
```
XuZhang99: we can place this unordered_map in the `namespace`, not in func `run()`.

btw, add `kimi_k2` to `deepseek_models`. 

maybe it's better change `deepseek_models` to `deepseek_like_model_set`.

`static` is useless, I think.

add check `FLAGS_backend!="dit"` or `FLAGS_backend="llm" || FLAGS_backend="vlm" || FLAGS_backend="rec"` before check whether model is deepseek-like or not.

xllm/xllm.cpp
```diff
@@ -126,6 +139,19 @@ int run() {
...
+      "deepseek_v2",
+      "deepseek_v3",
+      "deepseek_mtp",
+  };
+  if (deepseek_models.find(model_type) != deepseek_models.end()) {
+    FLAGS_enable_mla = true;
+  } else {
+    FLAGS_enable_mla = false;
```
XuZhang99: line 152 is useless, because its default value is false.

https://github.com/jd-opensource/xllm/blob/ca4be88a37896df8775b030db308578d8fd9332d/xllm/core/common/global_flags.cpp#L79-L81

xllm/xllm.cpp
```diff
@@ -45,6 +45,24 @@ void shutdown_handler(int signal) {
   exit(1);
 }
 
+std::string get_model_type(const std::filesystem::path& model_path) {
+  JsonReader reader;
+  // for llm, vlm and rec models, the config.json file is in the model path
+  std::filesystem::path config_json_path = model_path / "config.json";
+
+  if (std::filesystem::exists(config_json_path)) {
```
XuZhang99, line 61: no need to check `std::filesystem::exists(config_json_path)`, and remove `if`.

xllm/xllm.cpp
liuly0322: I think we should also modify the default value of `enable_mla`, or simply remove it.

1. If the auto choose logic is not always the optimal/right, we should allow user to override the default config, like the `backend` flag
https://github.com/jd-opensource/xllm/blob/ca4be88a37896df8775b030db308578d8fd9332d/xllm/core/common/global_flags.cpp#L59-L63
2. otherwise, the flag is useless, we should remove it.

RobbieLeung: Users don't always set the `enable_mla` parameter correctly, so it's reasonable for model contributors to control the auto-selection logic.

xllm/xllm.cpp
RobbieLeung: Optimize it by always using `get_model_type()` for this check logic.

xllm/xllm.cpp
liuly0322: > logic

Then why should we keep `FLAGS_enable_mla`? It is now fully controlled by model type.

RobbieLeung: > Then why should we keep `FLAGS_enable_mla`? It is now fully controlled by model type.

I will remove it later.

liuly0322 (approved):

PR #542, opened 2025-12-14
refactor: add common attention metadata for all cuda-like devices.

RobbieLeung:


xllm/core/layers/cuda/attention.cpp
```diff
@@ -144,10 +96,6 @@ std::tuple<torch::Tensor, std::optional<torch::Tensor>> AttentionImpl::forward(
     attention_params.k_cache = k_cache;
     attention_params.v_cache = v_cache;
 
-    // for mlu
-    attention_params.block_table = attn_metadata.block_table;
-    attention_params.kv_seq_lens = attn_metadata.kv_seq_lens;
```
XuZhang99, line 149: keep line 149.

PR #544, opened 2025-12-15
bugfix: fix round fail of LatencySchedule UT in a3.

weizhehuang0827:
bugfix: fix round fail of LatencySchedule UT in a3.


RobbieLeung (approved):

yq33victor (approved):
LGTM

PR #545, opened 2025-12-15
feat: add service guard to reject invalued requests.

RobbieLeung:


xllm/api_service/api_service.cpp
```diff
@@ -225,28 +269,17 @@ void APIService::Embeddings(::google::protobuf::RpcController* controller,
...
-  xllm::ClosureGuard done_guard(
-      done,
-      std::bind(request_in_metric, nullptr),
-      std::bind(request_out_metric, (void*)controller));
-  if (!request || !response || !controller) {
-    LOG(ERROR) << "brpc request | respose | controller is null";
-    return;
-  }
-  auto arena = response->GetArena();
+void EmbeddingImpl(std::unique_ptr<Service>& embedding_service_impl_,
```
yq33victor: nit: modify function name to `handle_embedding` or something like this. `EmbeddingImpl` seems like a class name.

xllm/api_service/api_service.cpp
```diff
@@ -38,49 +38,78 @@ limitations under the License.
 #include "xllm_metrics.h"
 namespace xllm {
 
+namespace {
+std::string unsupported_reason = "";
```
yq33victor: nit: maybe make it as a class member is better because it was modified in APIService ctor. 🥇  :)

xllm/api_service/api_service.cpp
RobbieLeung: Done. Rename `EmbeddingImpl` to `handle_embedding_request` and `ChatCompletionsImpl` to `handle_chat_request`.

PR #551, opened 2025-12-16
bugfix: support disagg PD for MTP.

RobbieLeung:


xllm/core/framework/kv_cache/spec_kv_cache_transfer.cpp
```diff
@@ -161,80 +168,55 @@ bool SpecKVCacheTransfer::pull_kv_blocks(
...
-      CHECK_LDD_RET(llm_data_dist_->PushKvBlocks(v_cache_,
-                                                 v_cache_index,
-                                                 kv_info.src_blocks,
-                                                 kv_info.dst_blocks,
-                                                 ext_param));
-    }
-  }
-  return true;
+  return push_kv_blocks_internal(
+      merged_kv_infos, layer_synchronizer, num_layers_, k_cache_, v_cache_);
```
DragonFive: Does this means push_kv_blocks_spec will never called?

RobbieLeung: Here is a bug. I will fix it.

XuZhang99 (approved):

PR #559, opened 2025-12-18
bugfix: fix hang issue when enable schedule overlap.

yq33victor:


xllm/core/util/concurrent_queue.h
```diff
@@ -0,0 +1,126 @@
...
+      T value = std::move(queue_.front());
+      queue_.pop();
+      return value;
+    }
+
+    return absl::nullopt;
+  }
+
+  std::optional<T> try_pop() {
+    absl::MutexLock lock(&mutex_);
```
DragonFive, line 94: Can `absl::optional<T> ` in pop and `std::optional<T>` in try_pop consist?

xllm/core/util/threadpool.h
```diff
@@ -61,7 +61,7 @@ class ThreadPool final {
                      BlockingCounter* block_counter);
 
   std::vector<std::thread> threads_;
-  std::vector<moodycamel::BlockingConcurrentQueue<Runnable>> queues_;
+  std::vector<ConcurrentQueue<Runnable>> queues_;
```
DragonFive, line 64: if disable schedule overlap(often happen in rec), is there any performance diff  between BlockingConcurrentQueue and your mutex implement?

DongheJin (dismissed):

xllm/core/util/threadpool.h
yq33victor: > if disable schedule overlap(often happen in rec), is there any performance diff between BlockingConcurrentQueue and your mutex implement?

This won't affect performance, the thread pool in xLLM typically has only one thread. schedule_overlap will reduce 10% E2E cost, the impact on performance is even greater.

xllm/core/util/concurrent_queue.h
yq33victor: > Can `absl::optional<T> ` in pop and `std::optional<T>` in try_pop consist?

fixed

DragonFive (approved):

PR #569, opened 2025-12-18
bugfix: resolve precision loss caused by prefix cache.

DongheJin:


yq33victor (approved):
LGTM:

we need to cherry-pick all bugfix to main branch from release/v0.7.0 later.

xllm/core/layers/base_layer.h
```diff
@@ -130,6 +130,7 @@ class BaseLayer : public torch::nn::Module {
   std::string name_;
   torch::ScalarType dtype_;
   std::vector<int> placeholder_vec_;
+  std::vector<int> placeholder_vec_zero_;
```
DragonFive: better change to int32_t because you use const_cast<int32_t*> in other place.

DongheJin: ok

PR #571, opened 2025-12-19
feat:enable arena for non-stream svcs and fix embedding/imageGeneration/rerank svcs mem leak.

xiao-yu-chen:


xllm/api_service/stream_call.h
```diff
@@ -160,4 +160,13 @@ class StreamCall : public Call {
   int connection_status_ = 0;
 };
 
+template <typename T>
+struct is_stream_call : std::false_type {};
+
+template <typename... Args>
+struct is_stream_call<StreamCall<Args...>> : std::true_type {};
```
DragonFive, line 167: I think here  is not good. CompletionCall and chatCall are alias of streamcall, so it will also be true.
`using CompletionCall = StreamCall`

xiao-yu-chen: yes, the implementation logic itself is required to be true， because they also provide streaming output to users

RobbieLeung (approved):

PR #577, opened 2025-12-19
refactor: standardize code style in function call logic.

yingxudeng:


xllm/function_call/base_format_detector.cpp
```diff
@@ -111,7 +111,7 @@ std::vector<ToolCallItem> BaseFormatDetector::parse_base_json(
   return results;
 }
 
-int BaseFormatDetector::_ends_with_partial_token(
+int BaseFormatDetector::ends_with_partial_token(
```
XuZhang99: specify `int` is `int32_t` or `int64_t`.

xllm/function_call/base_format_detector.h
```diff
@@ -69,8 +69,8 @@ class BaseFormatDetector {
   std::string eot_token_;
   std::string tool_call_separator_;
 
-  int _ends_with_partial_token(const std::string& buffer,
-                               const std::string& bot_token) const;
+  int ends_with_partial_token(const std::string& buffer,
```
XuZhang99: ditto.

change all `int` in function_call dir.

PR #585, opened 2025-12-23
refactor: optimize debugging logs and fix some parameter setting bugs.

magicheng0816:


xllm/cc_api/types.h
```diff
@@ -37,7 +37,7 @@ struct XLLM_CAPI_EXPORT XLLM_InitLLMOptions {
   // Whether to enable multi-head latent attention
   bool enable_mla = false;
 
-  bool disable_chunked_prefill = false;
+  bool disable_chunked_prefill = true;
 
   bool disable_prefix_cache = false;
```
XuZhang99: why not use `enable_chunked_prefill` and `enable_prefix_cache` to align with `global_flags.h` in xllm.

XuZhang99:
I think commit msg should be `refactor` not `feat`.

xllm/cc_api/llm.cpp
```diff
@@ -18,13 +18,38 @@ limitations under the License.
 #include <folly/Unit.h>
 #include <folly/experimental/coro/Timeout.h>
 #include <folly/futures/Future.h>
+#include <gflags/gflags.h>
```
DragonFive: Does here need this header?

magicheng0816: done

xllm/cc_api/types.h
magicheng0816: just align with llm.py

magicheng0816: done

DragonFive (approved):

PR #596, opened 2025-12-24
feat: implement prefix cache for decoder.

Kang-Meng:


xllm/core/framework/block/block_manager_pool.cpp
```diff
@@ -165,6 +165,38 @@ std::vector<Block> BlockManagerPool::allocate(size_t num_tokens,
   return block_managers_[dp_rank]->allocate(num_blocks_needed);
 }
 
+bool BlockManagerPool::try_allocate(Sequence* sequence) {
```
JimHsiung, line 168: Why not directly use allocate(Sequence* sequence)?

JimHsiung (approved):

PR #608, opened 2025-12-26
bugfix: correct max_context_len and max_batch_size calc for disagg PD…

czynb666:
… profiling. reserve one block for swap allocation.


czynb666:
block 0 reserved for padding: see the constructor of BlockManagerImpl

PR #617, opened 2025-12-29
refactor: remove useless npu rotary embedding.

RobbieLeung:


third_party/xllm_ops
XuZhang99, line 1: why change xllm_ops?

RobbieLeung: Done.

yingxudeng (approved):

PR #652, opened 2026-01-05
bugfix: fix MiMo-VL weights loading crash on NPU device.

wly-115:
bugfix: fix MiMo-VL weights loading crash on NPU device.


yq33victor (approved):
LGTM

PR #643, opened 2026-01-05
feat: add chat support for backend=rec for rec framwork in stretch goals[1/N].

DragonFive:
This PR is a follow-up to #619 and continues the generative recommendation work by adding
OpenAI-compatible chat completions API support for the rec framework.

## What's included

- **ChatServiceImpl RecMaster support**:
  - New constructor accepting `RecMaster*` for rec backend routing
  - `process_rec_chat_request()` handling `/v1/chat/completions` for LlmRec models
  - Automatic backend routing: RecMaster requests bypass LLMMaster path

- **RecMaster chat interface** (`handle_request(messages, ...)`):
  - Chat template application via `JinjaChatTemplate` for Qwen3
  - Message-to-prompt conversion with tool_calls support
  - Streaming and non-streaming response modes

- **Dynamic max_tokens support**:
  - `get_max_steps_from_batch()` reads `max_tokens` from `StoppingChecker`
  - Replaces fixed `kRecDecodeSteps` when stopping checker is present
  - Falls back to `kRecDecodeSteps` for OneRec compatibility

- **Batch API enhancement**:
  - `Batch::get_sequences()` moved to public for RecEngine access
  - Handles both `sequences_` (LlmRec) and `sequence_groups_` (OneRec) scenarios

## Usage

With `backend=rec`, LlmRec models now support the standard chat completions API:

```bash
curl http://localhost:8080/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{
    "model": "Qwen3-8B",
    "messages": [{"role": "user", "content": "Hello!"}],
    "max_tokens": 100,
    "stream": true
  }'
```

## Restrictions

- Chat completions API is **only supported for kLlmRec** models
- OneRec models return `INVALID_ARGUMENT` error for chat requests

## Testing

- [x] `backend=rec` + `/v1/chat/completions` (Qwen3-8B) — works
- [x] `backend=rec` + streaming mode — works
- [x] `backend=rec` + non-streaming mode — works
- [x] `backend=llm` + `/v1/chat/completions` — unchanged

## Roadmap checklist

- [x] Rec tokenizer #317
- [x] Rec decoder constrain #480
- [x] Rec proto and service #440
- [x] Rec engine master #557
- [x] Scheduler #557
- [x] RecType plumbing + rec batch input builder layering #565
- [x] Rec worker #567
- [x] Qwen3/LlmRec in rec framework #619
- [ ] LlmRec with mm_data
- [ ] Pure NPU inference path
[...]

PR #671, opened 2026-01-07
bugfix: fix kLlmRec KV cache allocation insufficient in rec FixedStepsSchduler.

DragonFive:
This PR is a follow-up bugfix to the chat completions API support added in the previous PR #643 . It fixes a KV cache allocation issue that caused assertion failures during LlmRec inference.

  ## Problem

  When running LlmRec models with the chat completions API, the following assertion failure occurred:

  batch_input_builder.cpp:283] Check failed: sequence->kv_state().current_max_tokens_capacity() >= seq_len

  The root cause was that `FixedStepsScheduler::LlmRecSchedulerPipeline::allocate_kv_cache()` was not properly allocating KV cache capacity for the full sequence length (prefill tokens + max generated tokens).

  ## What's fixed

  - **KV cache allocation in LlmRecSchedulerPipeline**:
    - Now correctly calculates required capacity: `num_tokens + max_generated_tokens`
    - Uses `KVCacheManager::allocate()` instead of `BlockManagerPool` for proper capacity tracking
    - Ensures `current_max_tokens_capacity() >= seq_len` check passes

  ## Files changed

  - `xllm/core/scheduler/fixed_steps_scheduler.cpp`
  - `xllm/core/scheduler/fixed_steps_scheduler.h`

  ## Testing

  - [x] `backend=rec` + `/v1/chat/completions` (Qwen3-8B) — no longer crashes
  - [x] KV cache capacity assertion passes

### DIFF
--- a/xllm/core/scheduler/fixed_steps_scheduler.cpp
+++ b/xllm/core/scheduler/fixed_steps_scheduler.cpp
@@ -28,6 +28,7 @@ limitations under the License.
 #include <memory>
 
 #include "common/metrics.h"
+#include "common/rec_model_utils.h"
 #include "core/common/global_flags.h"
 #include "distributed_runtime/engine.h"
 #include "framework/batch/batch.h"
@@ -229,10 +230,31 @@ std::vector<Batch> FixedStepsScheduler::prepare_batch() {
   // Initialize pipeline before handle_prefill_requests for KV cache allocation
   if (!scheduler_pipeline_ && !waiting_priority_queue_.empty()) {
     auto rec_type = waiting_priority_queue_.top()->state().rec_type;
-    if (rec_type == RecType::kLlmRec) {
-      scheduler_pipeline_ = std::make_unique<LlmRecSchedulerPipeline>();
-    } else {
-      scheduler_pipeline_ = std::make_unique<OneRecSchedulerPipeline>();
+    // Convert RecType to RecModelKind, then get pipeline type
+    RecModelKind rec_model_kind = RecModelKind::kNone;
+    switch (rec_type) {
+      case RecType::kLlmRec:
+        rec_model_kind = RecModelKind::kLlmRec;
+        break;
+      case RecType::kOneRec:
+        rec_model_kind = RecModelKind::kOneRec;
+        break;
+      default:
+        rec_model_kind = RecModelKind::kNone;
+        break;
+    }
+    auto pipeline_type = get_rec_pipeline_type(rec_model_kind);
+    switch (pipeline_type) {
+      case RecPipelineType::kLlmRecDefault:
+      case RecPipelineType::kLlmRecWithMmData:
+        scheduler_pipeline_ = std::make_unique<LlmRecSchedulerPipeline>();
+        break;
+      case RecPipelineType::kOneRecDefault:
+        scheduler_pipeline_ = std::make_unique<OneRecSchedulerPipeline>();
+        break;
+      default:
+        LOG(FATAL) << "Unknown FixedStepsScheduler pipeline type: "
+                   << static_cast<int>(pipeline_type);
     }
   }
 
--- a/xllm/core/scheduler/fixed_steps_scheduler.h
+++ b/xllm/core/scheduler/fixed_steps_scheduler.h
@@ -28,7 +28,6 @@ limitations under the License.
 #include "common/types.h"
 #include "framework/batch/batch.h"
 #include "framework/batch/batch_factory.h"
-#include "framework/block/block_manager_pool.h"
 #include "framework/request/request.h"
 #include "framework/request/sequence.h"
 #include "runtime/xservice_client.h"


xllm/core/scheduler/fixed_steps_scheduler.cpp
```diff
@@ -346,6 +347,16 @@ std::vector<Batch> FixedStepsScheduler::LlmRecSchedulerPipeline::create_batches(
...
 }
 
+bool FixedStepsScheduler::LlmRecSchedulerPipeline::allocate_kv_cache(
+    KVCacheManager* kv_cache_manager,
+    Sequence* sequence) {
+  const size_t num_tokens = sequence->num_tokens();
+  const size_t max_generated_tokens =
+      sequence->stopping_checker()->get_max_generated_tokens();
+  return kv_cache_manager->allocate(sequence,
+                                    num_tokens + max_generated_tokens);
```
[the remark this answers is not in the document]
DragonFive: OK

xllm/core/scheduler/fixed_steps_scheduler.h
```diff
@@ -56,20 +57,29 @@ class FixedStepsScheduler final : public ContinuousScheduler {
...
+    bool allocate_kv_cache(KVCacheManager* kv_cache_manager,
+                           Sequence* sequence) override;
   };
 
   class OneRecSchedulerPipeline final : public SchedulerPipeline {
    public:
     std::vector<Batch> create_batches(FixedStepsScheduler& scheduler,
                                       BatchFactory* batch_factory) override;
     bool requires_kv_cache() const override { return false; }
+    bool allocate_kv_cache(KVCacheManager* /*kv_cache_manager*/,
```
XuZhang99, line 78: what is `bool allocate_kv_cache(KVCacheManager* /*kv_cache_manager*/, Sequence* /*sequence*/)`?

xllm/core/scheduler/fixed_steps_scheduler.h
```diff
@@ -28,6 +28,7 @@ limitations under the License.
 #include "common/types.h"
 #include "framework/batch/batch.h"
 #include "framework/batch/batch_factory.h"
+#include "framework/block/block_manager_pool.h"
```
XuZhang99: do we really need this header?

xllm/core/scheduler/fixed_steps_scheduler.cpp
```diff
@@ -224,6 +225,16 @@ std::vector<Batch> FixedStepsScheduler::prepare_batch() {
...
 
+  // Initialize pipeline before handle_prefill_requests for KV cache allocation
+  if (!scheduler_pipeline_ && !waiting_priority_queue_.empty()) {
+    auto rec_type = waiting_priority_queue_.top()->state().rec_type;
+    if (rec_type == RecType::kLlmRec) {
+      scheduler_pipeline_ = std::make_unique<LlmRecSchedulerPipeline>();
+    } else {
+      scheduler_pipeline_ = std::make_unique<OneRecSchedulerPipeline>();
+    }
+  }
```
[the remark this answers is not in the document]
DragonFive: The scheduler instance is owned by a single RecMaster, which is initialized with a fixed rec_type determined by the model type at startup. A RecMaster serving a kLlmRec model will never receive kOneRec requests, and vice versa. The "mixed rec_type" scenario you described is architecturally impossible — this is not a runtime assumption but a design invariant enforced at the Master layer. Adding redundant assertions for impossible states would be defensive programming against non-existent bugs.

magicheng0816 (dismissed):

xllm/core/scheduler/fixed_steps_scheduler.cpp
```diff
@@ -224,6 +226,16 @@ std::vector<Batch> FixedStepsScheduler::prepare_batch() {
   size_t remaining_seq_budget = std::max(options_.max_seqs_per_batch(), 1);
   size_t num_preempted_requests = 0;
 
+  // Initialize pipeline before handle_prefill_requests for KV cache allocation
+  if (!scheduler_pipeline_ && !waiting_priority_queue_.empty()) {
+    auto rec_type = waiting_priority_queue_.top()->state().rec_type;
+    if (rec_type == RecType::kLlmRec) {
+      scheduler_pipeline_ = std::make_unique<LlmRecSchedulerPipeline>();
+    } else {
```
XuZhang99: `else if (rec_type == RecType::kOneRec)` and add `else` for `RecType::kNone`.

xllm/core/scheduler/fixed_steps_scheduler.h
DragonFive: It's a function in rec, different pipeline s(for pure prefill onerec, qwen, and future xattn) have different behavior, so we need this. in pure prefill onerec, it does't need kv, so return true directly.

xllm/core/scheduler/fixed_steps_scheduler.cpp
DragonFive: Good idea!

xllm/core/scheduler/fixed_steps_scheduler.h
DragonFive: ok, i will delete it.

XuZhang99 (approved):

PR #675, opened 2026-01-08
bugfix: resolve Deepseek execution failure caused by invalid input.

DongheJin:


xllm/core/runtime/worker_impl.cpp
```diff
@@ -415,7 +413,13 @@ void WorkerImpl::prepare_work_before_execute(const ForwardInput& input,
                               device_,
                               dtype_,
                               is_prefill);
-    processed_input.input_params.dp_ep_padding_data = dp_ep_padding.build();
+    if (context_.get_parallel_args().dp_size() > 1 ||
+        context_.get_parallel_args().ep_size() > 1) {
+      processed_input.input_params.dp_ep_padding_data = dp_ep_padding.build();
+    } else if (FLAGS_enable_mla) {
+      processed_input.input_params.dp_ep_padding_data =
+          dp_ep_padding.build_placeholder();
```
JimHsiung: Constructing placeholders in worker impl is not a good idea. It might be better to move this down to the model decode layer where it is actually used.

DongheJin: done

PR #688, opened 2026-01-09
bugfix: fix rate_limit bug for stream and non-stream request in PD disagg and refactor some callback logics.

weizhehuang0827:


xllm/api_service/chat_service_impl.cpp
```diff
@@ -828,15 +829,19 @@ void ChatServiceImpl::process_async_impl(std::shared_ptr<ChatCall> call) {
...
+                                              request_id,
+                                              created_time,
+                                              model,
+                                              req_output,
+                                              tool_call_parser_format,
+                                              reasoning_parser_format,
+                                              is_force_reasoning,
+                                              json_tools);
+          } else {
+            return true;
```
XuZhang99: remove `else` and `return true` directly.

xllm/api_service/completion_service_impl.cpp
```diff
@@ -232,16 +232,24 @@ void CompletionServiceImpl::process_async_impl(
...
           return send_delta_to_client_brpc(
               call, include_usage, request_id, created_time, model, req_output);
+        } else {
+          if (!req_output.prefill_finished_in_pd) {
+            return send_result_to_client_brpc(
+                call, request_id, created_time, model, req_output);
+          } else {
+            // when request is non-stream, not to call
+            // send_result_to_client_brpc if request prefill ends in PD Disagg
+            return true;
```
XuZhang99: ditto.

xllm/core/framework/request/request.cpp
```diff
@@ -142,7 +142,8 @@ void Request::handle_last_token() {
 }
 
 RequestOutput Request::generate_output(const Tokenizer& tokenizer,
-                                       ThreadPool* thread_pool) {
+                                       ThreadPool* thread_pool,
+                                       bool prefill_finished_in_pd) {
```
XuZhang99: what's the meaning of `prefill_finished_in_pd`? I don't think it's a good name.

weizhehuang0827:
```
// whether the request's prefill stage is finished for PD Disagg.
  // If in default mode, it should be always false.
  bool prefill_finished_in_pd = false;
```
explained in request_output.h

JimHsiung (approved):

PR #722, opened 2026-01-14
bugfix: fix streaming tool call missing function name for GLM-4.7

QwertyJack:
## Summary

- Fix streaming tool call not returning function name when stream mode is enabled for GLM-4.7 model
- When the buffer contained `<tool_call>` but the function name hadn't arrived yet, the non-greedy regex would match an empty string,
causing the tool call to be sent with an empty name

## Root Cause
In `parse_streaming_increment()`, when `func_name` was empty, the code would:
1. Log a warning but still send the tool call with an empty name
2. Mark `current_tool_name_sent_ = true`, preventing the name from ever being sent later

## Fix
Return early and wait for more streaming data when `func_name` is empty, instead of sending an empty function name.

## Test plan
- [x] Build the project: `python setup.py build`
- [x] Start the server: `cd /workdir && ./run.sh`
- [x] Run the test with streaming enabled: `python test.py`
- [x] Verify the warning `func_name should not be empty` no longer appears in logs
- [x] Verify the streaming response includes the function name in the tool call

### DIFF
--- a/xllm/function_call/glm47_detector.cpp
+++ b/xllm/function_call/glm47_detector.cpp
@@ -519,8 +519,13 @@ StreamingParseResult Glm47Detector::parse_streaming_increment(
 
       // Send tool name first if not sent yet
       if (!current_tool_name_sent_) {
-        if (func_name.empty()) {
-          LOG(WARNING) << "func_name should not be empty";
+        // Only send function name when we're sure it's complete:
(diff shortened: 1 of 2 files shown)


yq33victor:
@yingxudeng help to review this PR.

jayggh:
LGTM

XuZhang99 (approved):

PR #727, opened 2026-01-14
bugfix: return HTTP 400 instead of crashing when Content-Length header is missing.

QwertyJack:
## Summary

- Fix server crash when HTTP request has missing, invalid, or excessively large `Content-Length` header
- Return proper HTTP error responses instead of crashing the process
- Add payload size limit (10MB) to prevent DoS via memory exhaustion

## Problem

The server would crash in multiple scenarios:
1. Missing `Content-Length` header → `LOG(FATAL)` called
2. Invalid `Content-Length` value (non-numeric) → unhandled `std::stoul` exception
3. Excessively large `Content-Length` → memory exhaustion DoS

## Solution

- Change `GetJsonContentLength` to return `std::optional<size_t>`
- Add try-catch around `std::stoul` to handle invalid values
- Add 10MB size limit check before memory allocation
- Return HTTP 400 for missing/invalid headers
- Return HTTP 413 for oversized payloads

## Test plan

- Build: `python setup.py build`
- Start server: `(cd /workdir; ./run.sh)`
- Test missing header → verify HTTP 400
- Test invalid header (Content-Length: abc) → verify HTTP 400
- Test oversized header (Content-Length: 999999999999) → verify HTTP 413
- Test normal requests still work correctly

### DIFF
--- a/xllm/api_service/api_service.cpp
+++ b/xllm/api_service/api_service.cpp
@@ -45,12 +45,6 @@ namespace xllm {
 
 namespace {
 
-// Max request body size for API endpoints. 10MB accommodates typical LLM
-// requests including vision models with base64-encoded images. For larger
-// payloads, consider adjusting brpc's -max_body_size flag (default 64MB) which
-// acts as the first line of defense at the framework level.
-constexpr size_t kMaxBodySize = 10 * 1024 * 1024;  // 10MB
-
 std::optional<size_t> GetJsonContentLength(const brpc::Controller* ctrl) {
   const auto* header_val = ctrl->http_request().GetHeader(kInferContentLength);
   if (header_val == nullptr) {
@@ -79,9 +73,9 @@ std::optional<size_t> GetJsonContentLength(const brpc::Controller* ctrl) {
   return length;
 }
 
-// Validates Content-Length header and checks against max body size.
-// Returns validated content length, or nullopt if validation fails
-// (also sets appropriate HTTP error response).
+// Validates Content-Length header. Returns validated content length, or
+// nullopt if validation fails (also sets appropriate HTTP error response).
+// Note: brpc's -max_body_size flag provides payload size protection.
 std::optional<size_t> ValidateContentLength(brpc::Controller* ctrl) {
   auto content_len_opt = GetJsonContentLength(ctrl);
   if (!content_len_opt.has_value()) {
@@ -89,17 +83,7 @@ std::optional<size_t> ValidateContentLength(brpc::Controller* ctrl) {
     ctrl->http_response().set_status_code(400);
     return std::nullopt;
   }
-
-  auto content_len = content_len_opt.value();
-  if (content_len > kMaxBodySize) {
-    ctrl->SetFailed("Payload too large");
-    ctrl->http_response().set_status_code(413);
-    LOG(WARNING) << "Request body size " << content_len << " exceeds limit of "
-                 << kMaxBodySize;
-    return std::nullopt;
-  }
-
-  return content_len;
+  return content_len_opt.value();
 }
 
 template <typename Call>


xllm/api_service/api_service.cpp
```diff
@@ -136,17 +141,36 @@ void APIService::CompletionsHttp(::google::protobuf::RpcController* controller,
...
   }
 
+  auto ctrl = reinterpret_cast<brpc::Controller*>(controller);
+
+  auto content_len_opt = GetJsonContentLength(ctrl);
+  if (!content_len_opt.has_value()) {
+    ctrl->SetFailed("Missing or invalid Content-Length header");
+    ctrl->http_response().set_status_code(400);
+    LOG(WARNING) << "Missing or invalid Content-Length header, returning 400";
+    return;
```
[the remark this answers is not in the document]
QwertyJack: Thanks for the review! However, this is a false positive.

Looking at line 135-138, `xllm::ClosureGuard done_guard(done, ...)` is created **before** any early returns. The `ClosureGuard` destructor automatically calls `done->Run()` when it goes out of scope - including on early returns. This is the idiomatic RAII pattern for handling RPC closures.

The same pattern is used consistently across all the HTTP handlers in this file (`CompletionsHttp`, `ImageGenerationHttp`, `RerankHttp`, etc.) - they all create the guard at the beginning of the function.

xllm/api_service/api_service.cpp
```diff
@@ -136,17 +141,36 @@ void APIService::CompletionsHttp(::google::protobuf::RpcController* controller,
...
   auto resp_pb =
       google::protobuf::Arena::CreateMessage<proto::CompletionResponse>(arena);
 
-  auto ctrl = reinterpret_cast<brpc::Controller*>(controller);
   std::string error;
   json2pb::Json2PbOptions options;
-  butil::IOBuf& buf = ctrl->request_attachment();
-  butil::IOBufAsZeroCopyInputStream iobuf_stream(buf);
+  butil::IOBuf body_buf;
+  ctrl->request_attachment().cutn(&body_buf, content_len);
```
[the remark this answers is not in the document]
QwertyJack: This is a theoretical concern but not a practical vulnerability in our brpc setup.

In brpc, `request_attachment()` returns the **actual received data**, not what `Content-Length` claims. The HTTP parsing and connection handling (including read timeouts) are managed at the brpc framework layer, not at the application layer.

If an attacker sends `Content-Length: 10MB` but only 100 bytes of data:
- brpc's HTTP parser will wait for the remaining data with its configured timeout
- If timeout occurs, the connection is closed by brpc before our handler is even called
- `cutn` operates on already-received data in the IOBuf, it doesn't perform network I/O

The real protection against slow-loris style attacks is configured at the brpc server level (`idle_timeout_s`, `max_body_size`, etc.), which is the appropriate layer for such concerns.

That said, I agree the validation logic could be extracted into a helper function for maintainability - I'll consider that as a follow-up refactoring.

xllm/api_service/api_service.cpp
```diff
@@ -172,20 +196,28 @@ void APIService::ChatCompletions(::google::protobuf::RpcController* controller,
...
 
-  LOG(FATAL) << "Content-Length header is missing.";
-  return (size_t)-1L;
+  try {
+    const long long length = std::stoll(*header_val);
+    if (length < 0) {
+      LOG(WARNING) << "Invalid negative Content-Length header: " << *header_val;
+      return std::nullopt;
+    }
+    return static_cast<size_t>(length);
```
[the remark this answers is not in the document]
QwertyJack: Good catch! Fixed in the latest commit.

Added a check for `std::numeric_limits<size_t>::max()` before the cast:

```cpp
if (static_cast<unsigned long long>(length) >
    std::numeric_limits<size_t>::max()) {
  LOG(WARNING) << "Content-Length exceeds size_t max: " << *header_val;
  return std::nullopt;
}
```

This ensures proper handling on 32-bit systems where `size_t` is smaller than `long long`.

xllm/api_service/api_service.cpp
```diff
@@ -136,17 +142,36 @@ void APIService::CompletionsHttp(::google::protobuf::RpcController* controller,
...
+  }
+  auto content_len = content_len_opt.value();
+
+  if (content_len > kMaxBodySize) {
+    ctrl->SetFailed("Payload too large");
+    ctrl->http_response().set_status_code(413);
+    LOG(WARNING) << "Request body size " << content_len << " exceeds limit of "
+                 << kMaxBodySize;
+    return;
+  }
```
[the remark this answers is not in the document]
QwertyJack: Good suggestion! Fixed in the latest commit.

Created a `ValidateContentLength` helper function that combines the validation logic:

```cpp
std::optional<size_t> ValidateContentLength(brpc::Controller* ctrl) {
  auto content_len_opt = GetJsonContentLength(ctrl);
  if (!content_len_opt.has_value()) {
    ctrl->SetFailed("Missing or invalid Content-Length header");
    ctrl->http_response().set_status_code(400);
    LOG(WARNING) << "Missing or invalid Content-Length header, returning 400";
    return std::nullopt;
  }

  auto content_len = content_len_opt.value();
  if (content_len > kMaxBodySize) {
    ctrl->SetFailed("Payload too large");
    ctrl->http_response().set_status_code(413);
    LOG(WARNING) << "Request body size " << content_len << " exceeds limit of "
                 << kMaxBodySize;
    return std::nullopt;
  }

  return content_len;
}
```

All 7 HTTP handlers now use this helper, reducing ~100 lines of duplication.

xllm/api_service/api_service.cpp
```diff
@@ -172,20 +186,57 @@ void APIService::ChatCompletions(::google::protobuf::RpcController* controller,
...
+        std::numeric_limits<size_t>::max()) {
+      LOG(WARNING) << "Content-Length exceeds size_t max: " << *header_val;
+      return std::nullopt;
+    }
+    return static_cast<size_t>(length);
+  } catch (const std::exception& e) {
+    LOG(WARNING) << "Invalid Content-Length header '" << *header_val
+                 << "': " << e.what();
+    return std::nullopt;
   }
```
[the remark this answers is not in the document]
QwertyJack: Thanks for the suggestion, but I don't think this is a practical concern in our context.

**brpc validates first**: The HTTP layer in brpc parses and validates the `Content-Length` header before our handler is called. Our code only sees headers that brpc has already accepted. If brpc were to accept a malformed header like "123xyz", that would be a bug in brpc's HTTP implementation.

**Request smuggling requires parser disagreement**: HTTP request smuggling exploits differences between how a front-end proxy and back-end server parse headers. But our code isn't the HTTP parser - brpc is. The `std::stoll` parsing here is a second layer of defense, not the primary HTTP parser.

**Current code is already defensive**: `std::stoll` throws exceptions for completely invalid strings (handled by our catch block). The only edge case is "123xyz" style strings, which are unlikely to pass brpc's HTTP parser in the first place.

Adding a manual digit-checking loop would add complexity for a scenario that's unlikely to occur in practice given brpc's HTTP validation.

xllm/api_service/api_service.cpp
```diff
@@ -200,7 +251,12 @@ void ChatCompletionsImpl(std::unique_ptr<Service>& service,
   auto resp_pb =
       google::protobuf::Arena::CreateMessage<typename ChatCall::ResType>(arena);
 
-  auto content_len = GetJsonContentLength(ctrl);
+  auto content_len_opt = ValidateContentLength(ctrl);
+  if (!content_len_opt.has_value()) {
+    return;
+  }
+  auto content_len = content_len_opt.value();
```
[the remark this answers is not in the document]
QwertyJack: Good catch! Fixed in the latest commit.

Refactored `ChatCompletionsImpl` to use the zero-copy approach consistent with other handlers:

```cpp
std::string error;
json2pb::Json2PbOptions options;
butil::IOBuf body_buf;
ctrl->request_attachment().cutn(&body_buf, content_len);
butil::IOBufAsZeroCopyInputStream iobuf_stream(body_buf);
auto st = json2pb::JsonToProtoMessage(&iobuf_stream, req_pb, options, &error);
```

Also removed the now-unused `<google/protobuf/util/json_util.h>` include.

xllm/api_service/api_service.cpp
```diff
@@ -200,17 +250,21 @@ void ChatCompletionsImpl(std::unique_ptr<Service>& service,
...
-    ctrl->SetFailed(status.ToString());
-    LOG(ERROR) << "parse json to proto failed: " << status.ToString();
+  auto content_len_opt = ValidateContentLength(ctrl);
+  if (!content_len_opt.has_value()) {
+    return;
+  }
+  auto content_len = content_len_opt.value();
+
+  std::string error;
+  json2pb::Json2PbOptions options;
```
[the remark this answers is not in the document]
QwertyJack: Thanks for the review, but this is actually not a regression.

Looking at the `json2pb::Json2PbOptions` struct in `third_party/brpc/src/json2pb/json_to_pb.h`, there is **no `ignore_unknown_fields` option**. The only options are:
- `base64_to_bytes`
- `array_to_single_repeated`
- `allow_remaining_bytes_after_parsing`

Furthermore, checking the `json2pb` implementation, it iterates over **protobuf fields** and looks for matching JSON members (not the other way around). Unknown JSON fields are simply **ignored by design** because they're never looked up.

This is the same behavior as `google::protobuf::util::JsonParseOptions` with `ignore_unknown_fields = true` - both implementations ignore extra JSON fields that don't map to protobuf fields.

xllm/api_service/api_service.cpp
```diff
@@ -136,17 +194,25 @@ void APIService::CompletionsHttp(::google::protobuf::RpcController* controller,
...
   std::string error;
   json2pb::Json2PbOptions options;
-  butil::IOBuf& buf = ctrl->request_attachment();
-  butil::IOBufAsZeroCopyInputStream iobuf_stream(buf);
+  butil::IOBuf body_buf;
+  ctrl->request_attachment().cutn(&body_buf, content_len);
+  butil::IOBufAsZeroCopyInputStream iobuf_stream(body_buf);
   auto st = json2pb::JsonToProtoMessage(&iobuf_stream, req_pb, options, &error);
   if (!st) {
     ctrl->SetFailed(error);
```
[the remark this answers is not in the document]
QwertyJack: Thanks for the suggestion! The validation logic has already been consolidated into the `ValidateContentLength` helper function (lines 80-98).

The remaining pattern (JSON parsing with `json2pb`) is intentionally kept inline because:
1. Each handler has slightly different protobuf message types (`CompletionRequest`, `ChatRequest`, `EmbeddingRequest`, etc.)
2. The template function `ChatCompletionsImpl` already abstracts the pattern for chat-related handlers
3. Further abstraction would require complex template machinery for marginal benefit

The current structure balances code reuse (via `ValidateContentLength`) with readability and type safety.

xllm/api_service/api_service.cpp
```diff
@@ -200,7 +206,14 @@ void ChatCompletionsImpl(std::unique_ptr<Service>& service,
...
+  auto content_len_opt = GetJsonContentLength(ctrl);
+  if (!content_len_opt.has_value()) {
+    ctrl->SetFailed("Content-Length header is missing");
+    ctrl->http_response().set_status_code(400);
+    LOG(WARNING) << "Content-Length header is missing, returning 400";
+    return;
+  }
+  auto content_len = content_len_opt.value();
   std::string attachment;
   ctrl->request_attachment().copy_to(&attachment, content_len, 0);
```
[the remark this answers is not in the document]
QwertyJack: Fixed in the latest commit. Added `kMaxBodySize` check in `ValidateContentLength` helper function (lines 89-96) which returns HTTP 413 for oversized payloads.

xllm/api_service/api_service.cpp
```diff
@@ -200,7 +206,14 @@ void ChatCompletionsImpl(std::unique_ptr<Service>& service,
   auto resp_pb =
       google::protobuf::Arena::CreateMessage<typename ChatCall::ResType>(arena);
 
-  auto content_len = GetJsonContentLength(ctrl);
+  auto content_len_opt = GetJsonContentLength(ctrl);
+  if (!content_len_opt.has_value()) {
+    ctrl->SetFailed("Content-Length header is missing");
+    ctrl->http_response().set_status_code(400);
+    LOG(WARNING) << "Content-Length header is missing, returning 400";
+    return;
```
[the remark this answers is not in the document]
QwertyJack: Fixed in the latest commit. The error message now uses 'Missing or invalid Content-Length header' in the `ValidateContentLength` helper function (line 83).

xllm/api_service/api_service.cpp
```diff
@@ -39,6 +41,62 @@ limitations under the License.
...
+    header_val = ctrl->http_request().GetHeader(kContentLength);
+  }
+
+  if (header_val == nullptr) {
+    LOG(WARNING) << "Content-Length header is missing";
+    return std::nullopt;
+  }
+
+  try {
+    const long long length = std::stoll(*header_val);
```
yq33victor: we don't need `try-catch` here.

xllm/api_service/api_service.cpp
```diff
@@ -39,6 +41,62 @@ limitations under the License.
 namespace xllm {
 
 namespace {
+
+constexpr size_t kMaxBodySize = 10 * 1024 * 1024;  // 10MB
```
yq33victor: nit: why 10MB here , Are there some bases?

xllm/api_service/api_service.cpp
```diff
@@ -136,17 +194,25 @@ void APIService::CompletionsHttp(::google::protobuf::RpcController* controller,
     return;
   }
 
+  auto ctrl = reinterpret_cast<brpc::Controller*>(controller);
+
+  auto content_len_opt = ValidateContentLength(ctrl);
+  if (!content_len_opt.has_value()) {
+    return;
```
yq33victor, line 192: Add `ctrl->SetFailed(error);`

xllm/api_service/api_service.cpp
```diff
@@ -200,17 +250,21 @@ void ChatCompletionsImpl(std::unique_ptr<Service>& service,
...
-  google::protobuf::util::JsonParseOptions options;
-  options.ignore_unknown_fields = true;
-  auto status =
-      google::protobuf::util::JsonStringToMessage(attachment, req_pb, options);
-  if (!status.ok()) {
-    ctrl->SetFailed(status.ToString());
-    LOG(ERROR) << "parse json to proto failed: " << status.ToString();
+  auto content_len_opt = ValidateContentLength(ctrl);
+  if (!content_len_opt.has_value()) {
+    return;
```
yq33victor, line 325: Add `ctrl->SetFailed(error);`

xllm/api_service/api_service.cpp
```diff
@@ -273,6 +327,15 @@ void handle_embedding_request(std::unique_ptr<Service>& embedding_service_impl_,
     LOG(ERROR) << "brpc request | respose | controller is null";
     return;
   }
+
+  auto ctrl = reinterpret_cast<brpc::Controller*>(controller);
+
+  auto content_len_opt = ValidateContentLength(ctrl);
+  if (!content_len_opt.has_value()) {
+    return;
```
yq33victor, line 415: Add `ctrl->SetFailed(error);`

xllm/api_service/api_service.cpp
```diff
@@ -340,6 +403,14 @@ void APIService::ImageGenerationHttp(
     return;
   }
 
+  auto ctrl = reinterpret_cast<brpc::Controller*>(controller);
+
+  auto content_len_opt = ValidateContentLength(ctrl);
+  if (!content_len_opt.has_value()) {
+    return;
```
yq33victor, line 490: Add `ctrl->SetFailed(error);`

xllm/api_service/api_service.cpp
```diff
@@ -385,17 +456,25 @@ void APIService::RerankHttp(::google::protobuf::RpcController* controller,
...
```
yq33victor, line 543: Add `ctrl->SetFailed(error);`

xllm/api_service/api_service.cpp
QwertyJack: Good catch! Fixed in f91c446.

xllm/api_service/api_service.cpp
QwertyJack: Added rationale in f91c446:

```cpp
// Max request body size for API endpoints. 10MB accommodates typical LLM requests
// including vision models with base64-encoded images. For larger payloads, consider
// adjusting brpc's -max_body_size flag (default 64MB) which acts as the first line
// of defense at the framework level.
```

xllm/api_service/api_service.cpp
QwertyJack: `ValidateContentLength` already calls `ctrl->SetFailed()` internally (see lines 85 and 92). The comment on the helper function documents this behavior:

```cpp
// Validates Content-Length header and checks against max body size.
// Returns validated content length, or nullopt if validation fails
// (also sets appropriate HTTP error response).
```

xllm/api_service/api_service.cpp
QwertyJack: Same as above - `ValidateContentLength` already calls `ctrl->SetFailed()` internally.

xllm/api_service/api_service.cpp
QwertyJack: Same as above - `ValidateContentLength` already calls `ctrl->SetFailed()` internally.

xllm/api_service/api_service.cpp
QwertyJack: Same as above - `ValidateContentLength` already calls `ctrl->SetFailed()` internally.

xllm/api_service/api_service.cpp
QwertyJack: Same as above - `ValidateContentLength` already calls `ctrl->SetFailed()` internally.

xllm/api_service/api_service.cpp
zhang-minchao: Based on the function name `ValidateContentLength`, it should only perform validation and should not include ctrl->SetFailed(). Moreover, it should return bool or Status.

xllm/api_service/api_service.cpp
yq33victor: My point is that we should reserve a configurable parameter and an interface here, it shouldn’t be a constant.
Defensive programming is needed here.

xllm/api_service/api_service.cpp
QwertyJack: You raise a fair point about naming purity. A function called `ValidateContentLength` arguably shouldn't have side effects.

However, I chose this design intentionally for DRY reasons: there are 6+ HTTP handlers calling this function. If it returned just `bool` or `Status`, every call site would need:

```cpp
auto result = ValidateContentLength(ctrl);
if (!result.ok()) {
  ctrl->SetFailed(result.message());
  ctrl->http_response().set_status_code(result.code());
  return;
}
```

That's duplicated boilerplate at every handler, and risks inconsistent error responses if someone forgets the status code or uses the wrong one.

The current approach centralizes the error handling logic in one place. The function signature (`brpc::Controller*` non-const) and the doc comment (lines 79-81) signal that it may modify the controller.

That said, if the name is misleading, I'm happy to rename it to something like `ValidateContentLengthOrFail()` or `EnsureValidContentLength()` to make the behavior explicit.

What do you think - is a rename sufficient, or do you prefer the pure validation approach?

zhang-minchao: I think the following code can be encapsulated as a macro, `RETURN_HTTP_FAILED_IF_ERROR`:

```cpp
if (!result.ok()) {
  ctrl->SetFailed(result.message());
  ctrl->http_response().set_status_code(result.code());
  return;
}
```
This way each call would look like
```cpp
RETURN_HTTP_FAILED_IF_ERROR(ValidateContentLength(ctrl));
```
This approach improves semantic clarity while eliminating redundant code.

QwertyJack: The `RETURN_HTTP_FAILED_IF_ERROR` macro pattern is elegant and widely used (similar to Abseil's `RETURN_IF_ERROR`).

However, there's a complication: our function needs to return the **content length value** on success, not just a status. The macro pattern requires a `Status` return type, but we need something like `StatusOr<size_t>`.

To make the macro work, we'd need to either:
1. Introduce a `StatusOr<T>` type (like `absl::StatusOr`)
2. Use an out-parameter: `Status ValidateContentLength(ctrl, &content_len)`
3. Split into separate validate and get functions

None of these exist in the codebase currently, and introducing them feels like scope creep for a bugfix PR.

**Proposal:** For this PR, I'll rename the function to `ValidateContentLengthOrFail()` to make the side-effect behavior explicit. The macro/StatusOr refactoring is a good idea but better suited for a separate cleanup PR.

Does that work for you?

xllm/api_service/api_service.cpp
QwertyJack: Removed the application-level size check entirely. brpc's `-max_body_size` flag (default 64MB) already provides this protection at the framework level, so the extra check was redundant.

PR #746, opened 2026-01-16
bugfix: fix mrope calculation in the multimodal situation.

shan-chen-feng:
xllm apply the repeat operation first "cat([cos, cos]), cat([sin, sin])", need to select the mrope part from both sections

### DIFF
--- a/xllm/models/llm/npu/qwen3_moe.h
+++ b/xllm/models/llm/npu/qwen3_moe.h
@@ -189,14 +189,21 @@ class Qwen3MoeModelImpl : public torch::nn::Module {
       auto apply = [this](torch::Tensor x) {
         // auto sections = mrope_section_;
         auto freqs_t = x[0].clone();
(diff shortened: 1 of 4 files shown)


xllm/models/llm/npu/qwen3_moe.h
```diff
@@ -189,14 +189,17 @@ class Qwen3MoeModelImpl : public torch::nn::Module {
       auto apply = [this](torch::Tensor x) {
         // auto sections = mrope_section_;
         auto freqs_t = x[0].clone();
+        int64_t half_dim = static_cast<int64_t>(freqs_t.size(-1) / 2);
```
XuZhang99: I remember `freqs_t.size(-1) / 2` is `head_dim`, so you need to double check. If so, rename `half_dim` to `head_dim`.

wly-115:
same as Qwen3 Dense.

PR #772, opened 2026-01-23
feat: support request cancelled by client in PD disagg mode.

weizhehuang0827:


xllm/core/runtime/xservice_client.h
```diff
@@ -51,7 +51,7 @@ class XServiceClient {
   std::vector<std::string> get_static_prefill_list();
 
   // response generation tokens to xllm service
-  void generations(const std::vector<RequestOutput>& outputs);
+  std::vector<bool> generations(const std::vector<RequestOutput>& outputs);
```
XuZhang99, line 54: `std::vector<bool>` is not good, use `std::vector<char>` instead.

xllm/core/scheduler/async_response_processor.cpp
```diff
@@ -33,15 +33,31 @@ namespace xllm {
...
+      role_(role.value_or(InstanceRole::DEFAULT)) {
   if (role_ == InstanceRole::DECODE || role_ == InstanceRole::MIX) {
-    enable_batch_response_ =
-        util::get_bool_env("ENABLE_PD_DECODE_BATCH_RESPONSE", true);
+    enable_batch_response_ = true;
+  }
+}
+
+void AsyncResponseProcessor::process_failed_request(
+    std::shared_ptr<Request> request,
```
XuZhang99, line 46: `const std::shared_ptr<Request>& request`

xllm/core/runtime/xservice_client.h
weizhehuang0827: why use char

PR #788, opened 2026-01-24
feat: add tpot and generation throughput to request logging.

QwertyJack:
## Summary
- Add TPOT (Time Per Output Token) metric to request logging
- Add generation throughput (tokens/s) metric to request logging
- Refactor log_statistic to compute metrics before logging

## Test plan
- [x] Build the project and verify compilation succeeds
- [x] Run the server and send test requests
- [x] Verify log output includes tpot and generation throughput fields

### DIFF
--- a/xllm/core/framework/request/request.cpp
+++ b/xllm/core/framework/request/request.cpp
@@ -71,6 +71,17 @@ void Request::log_statistic(double total_latency) {
   // log the request statistics
   int idx = 0;
   for (const auto& seq : sequences()) {
+    double ttft = seq->time_to_first_token_latency_seconds();
+    size_t gen_tokens = state_.enable_schedule_overlap
+                            ? seq->num_generated_tokens() - 1
+                            : seq->num_generated_tokens();
+    double tpot = 0.0;
+    double gen_speed = 0.0;
+    if (gen_tokens > 1 && total_latency > ttft && ttft > 0) {
+      const double generation_latency = total_latency - ttft;
+      tpot = (generation_latency * 1000.0) / (gen_tokens - 1);
+      gen_speed = gen_tokens / generation_latency;
+    }
     LOG(INFO) << "x-request-id: " << x_request_id_ << ", "
               << "x-request-time: " << x_request_time_ << ", "
               << "request_id: " << request_id_ << ", "
@@ -81,14 +92,11 @@ void Request::log_statistic(double total_latency) {
               << "finish_reason: "
               << seq->finish_reason().to_string().value_or("") << ", "
               << "prompt_tokens: " << seq->num_prompt_tokens() << ", "
-              << "generated_tokens: "
-              << (state_.enable_schedule_overlap
-                      ? seq->num_generated_tokens() - 1
-                      : seq->num_generated_tokens())
-              << ", " << std::fixed << std::setprecision(1)
-              << "ttft: " << seq->time_to_first_token_latency_seconds() * 1000
-              << "ms, "
-              << "total_latency: " << total_latency * 1000 << "ms";
+              << "generated_tokens: " << gen_tokens << ", " << std::fixed
+              << std::setprecision(1) << "ttft: " << ttft * 1000 << "ms, "
+              << "total_latency: " << total_latency * 1000 << "ms, "
+              << "avg tpot: " << tpot << "ms, "
+              << "generation speed: " << gen_speed << " tokens/s";
     // only log once when beam search is enabled
     if (check_beam_search()) {
       break;


xllm/core/framework/request/request.cpp
```diff
@@ -71,6 +71,16 @@ void Request::log_statistic(double total_latency) {
...
   for (const auto& seq : sequences()) {
+    double ttft = seq->time_to_first_token_latency_seconds();
+    size_t gen_tokens = state_.enable_schedule_overlap
+                            ? seq->num_generated_tokens() - 1
+                            : seq->num_generated_tokens();
+    double tpot = 0.0;
+    double throughput = 0.0;
+    if (gen_tokens > 0 && total_latency > ttft && ttft > 0) {
+      tpot = ((total_latency - ttft) * 1000.0) / gen_tokens;
+      throughput = 1000.0 / tpot;
```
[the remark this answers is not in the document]
QwertyJack: Good suggestion. Applied the change to use a direct `generation_latency` variable for better readability and to avoid the intermediate division.

xllm/core/framework/request/request.cpp
```diff
@@ -71,6 +71,17 @@ void Request::log_statistic(double total_latency) {
...
+    double ttft = seq->time_to_first_token_latency_seconds();
+    size_t gen_tokens = state_.enable_schedule_overlap
+                            ? seq->num_generated_tokens() - 1
+                            : seq->num_generated_tokens();
+    double tpot = 0.0;
+    double throughput = 0.0;
+    if (gen_tokens > 0 && total_latency > ttft && ttft > 0) {
+      const double generation_latency = total_latency - ttft;
+      tpot = (generation_latency * 1000.0) / gen_tokens;
+      throughput = gen_tokens / generation_latency;
```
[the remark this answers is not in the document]
QwertyJack: The existing guard `total_latency > ttft` ensures `generation_latency > 0`. In practice, latencies are measured in seconds with sub-millisecond precision, so an epsilon-level value is physically impossible. If such an anomaly did occur, logging an extreme throughput value would actually help identify the timing bug rather than silently masking it with 0. The added complexity isn't warranted here.

xllm/core/framework/request/request.cpp
```diff
@@ -71,6 +71,17 @@ void Request::log_statistic(double total_latency) {
...
   for (const auto& seq : sequences()) {
+    double ttft = seq->time_to_first_token_latency_seconds();
+    size_t gen_tokens = state_.enable_schedule_overlap
+                            ? seq->num_generated_tokens() - 1
+                            : seq->num_generated_tokens();
+    double tpot = 0.0;
+    double gen_speed = 0.0;
+    if (gen_tokens > 0 && total_latency > ttft && ttft > 0) {
+      const double generation_latency = total_latency - ttft;
+      tpot = (generation_latency * 1000.0) / gen_tokens;
```
weizhehuang0827: should be `gen_tokens-1`

xllm/core/framework/request/request.cpp
```diff
@@ -81,14 +92,11 @@ void Request::log_statistic(double total_latency) {
...
-                      ? seq->num_generated_tokens() - 1
-                      : seq->num_generated_tokens())
-              << ", " << std::fixed << std::setprecision(1)
-              << "ttft: " << seq->time_to_first_token_latency_seconds() * 1000
-              << "ms, "
-              << "total_latency: " << total_latency * 1000 << "ms";
+              << "generated_tokens: " << gen_tokens << ", " << std::fixed
+              << std::setprecision(1) << "ttft: " << ttft * 1000 << "ms, "
+              << "total_latency: " << total_latency * 1000 << "ms, "
+              << "tpot: " << tpot << "ms, "
```
XuZhang99: avg tpot, because there are many tokens generated in decode phase.

xllm/core/framework/request/request.cpp
QwertyJack: Fixed. TPOT measures the average time between consecutive tokens, so with N generated tokens there are N-1 intervals. Changed to `(gen_tokens - 1)` and guard condition to `gen_tokens > 1`.

xllm/core/framework/request/request.cpp
QwertyJack: Good point. Renamed the label to "avg tpot" since it's the average time per token across all decode steps.

weizhehuang0827: maybe no need. it reports single request's tpot.

XuZhang99: > maybe no need. it reports single request's tpot.

single request will generate many tokens in decode phase, and each token latency maybe be different.

PR #797, opened 2026-01-26
feat: support deepseek mla indexer fused_indexer_q/fused_indexer_k on mlu device.

a120092009:


xllm/core/layers/common/indexer.cpp
```diff
@@ -121,7 +122,8 @@ IndexerImpl::IndexerImpl(int64_t dim,
       index_topk_(index_topk),
       q_lora_rank_(q_lora_rank),
       rotary_emb_(rotary_emb),
-      softmax_scale_(std::pow(head_dim_, -0.5) * std::pow(n_heads_, -0.5)) {
+      softmax_scale_(std::pow(head_dim_, -0.5) * std::pow(n_heads_, -0.5)),
+      use_fused_qk_(use_fused_qk) {
```
XuZhang99: I think it's better to rename `use_fused_qk` to `enable_fused_qk`.

PR #823, opened 2026-01-28
bugfix: get suitable token budget that allocated for sequence when enable MTP with overlap.

RobbieLeung:


yq33victor (approved):
LGTM

yq33victor (approved):
LGTM

PR #827, opened 2026-01-28
bugfix: fix eager mode fallback crash in NPU decoder layers. (#826)

QwertyJack:
## Summary

- Fix eager mode fallback crash when `kv_max_seq_len` exceeds `max_seq_len_for_graph_mode`
- Provide `tensor_placeholder_` when `tiling_data` is not available in eager mode

## Root Cause

When the system falls back to eager mode, decoder nodes built with `enableAclGraphPagedAttention=true` expect a `tiling_data` input tensor. However, eager mode doesn't provide this tensor, causing ATB `Setup()` to fail with error code 8.

## Test plan

- [ ] Verify server starts normally with `--enable_graph=true`
- [ ] Test with requests where `kv_max_seq_len > max_seq_len_for_graph_mode` to trigger eager mode fallback
- [ ] Confirm no crash occurs during eager mode execution

Fixes #826

PR #829, opened 2026-01-29
bugfix: fix KV cache memory leak when prefix cache is enabled.

### LINKED ISSUE #828: KV cache memory leak when prefix cache is enabled
## Problem

When `enable_prefix_cache=true`, the server gradually refuses all requests with the error:
```
Request prompt is too long, no enough memory to schedule a single sequence.
```

The metrics show abnormal values:
```
kv_cache_utilization_perc 1.35506  (135.5%, exceeds 100%)
num_running_requests 0
num_running_sequences 0
num_waiting_requests 0
```

## Root Cause

In `BlockManagerPool::deallocate()`, `cache()` is called before `deallocate()`. This causes:

1. Sequence holds blocks (ref_count = 2)
2. `cache()` is called first → adds to prefix cache → ref_count = 3
3. `deallocate()` checks: `ref_count <= 2`? → FALSE → `num_used_blocks_` NOT decremented
4. Result: `num_used_blocks_` keeps growing, never decreases

The check in `BlockManagerImpl::deallocate()` expects ref_count to be 2 when the sequence is the only holder, but `cache()` has already incremented it.

## Impact

- **Trigger condition**: Only when `enable_prefix_cache=true`
- **Severity**: Server becomes unusable after processing enough requests
- **Scope**: All requests are rejected once utilization exceeds threshold

## Fix

Swap the order of `deallocate()` and `cache()` calls in `BlockManagerPool::deallocate()`.

This is safe because `deallocate()` only updates the `num_used_blocks_` counter - it doesn't actually free block memory. Block lifetime is managed by ref_count.

QwertyJack:
## Summary

- Fix KV cache memory leak that causes server to reject all requests when prefix cache is enabled
- Swap the order of `deallocate()` and `cache()` calls in `BlockManagerPool::deallocate()`

Fixes #828

## Problem

When `enable_prefix_cache=true`, `num_used_blocks_` is never decremented on request completion because:

1. `cache()` is called first, which inserts blocks into prefix cache and increments ref_count to 3
2. `deallocate()` checks `ref_count <= 2`, which is now FALSE
3. `num_used_blocks_` decrement is skipped

This causes `kv_cache_utilization` to exceed 100% and the scheduler to reject all new requests.

## Solution

Call `deallocate()` before `cache()` so the ref_count check sees the correct value (2) before the prefix cache adds its reference.

This is safe because `deallocate()` only updates the accounting counter - it doesn't free block memory. Block lifetime is managed by ref_count.

## Test plan

- [ ] Verify `kv_cache_utilization_perc` stays within 0-1 range after processing requests
- [ ] Verify `num_used_blocks` correctly decreases when requests complete
- [ ] Run existing test suite

### DIFF
--- a/xllm/core/framework/block/block_manager_pool.cpp
+++ b/xllm/core/framework/block/block_manager_pool.cpp
@@ -90,10 +90,13 @@ void BlockManagerPool::deallocate(std::vector<Sequence*>& sequences) {
 
 void BlockManagerPool::deallocate(Sequence* sequence) {
   DCHECK(sequence != nullptr);
-  // add blocks to the prefix cache
   int32_t dp_rank = get_dp_rank(sequence);
-  cache(sequence);
+  // deallocate must be called before cache() to correctly update
+  // num_used_blocks_ cache() increases ref_count, which would cause
+  // deallocate() to skip the decrement if called after
   block_managers_[dp_rank]->deallocate(sequence->kv_state().kv_blocks());
+  // add blocks to the prefix cache
+  cache(sequence);
   // release the blocks after prefix cache insertion
   sequence->reset();
 }


xllm/core/framework/block/block_manager_pool.cpp
```diff
@@ -90,10 +90,13 @@ void BlockManagerPool::deallocate(std::vector<Sequence*>& sequences) {
 
 void BlockManagerPool::deallocate(Sequence* sequence) {
   DCHECK(sequence != nullptr);
-  // add blocks to the prefix cache
   int32_t dp_rank = get_dp_rank(sequence);
-  cache(sequence);
+  // deallocate must be called before cache() to correctly update
+  // num_used_blocks_ cache() increases ref_count, which would cause
+  // deallocate() to skip the decrement if called after
```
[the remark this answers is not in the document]
QwertyJack: Thank you for the thorough analysis, but this concern does not apply to this codebase.

**The scheduler is single-threaded**: The `step()` function in `ContinuousScheduler` (and its subclasses) runs sequentially. `allocate_shared()` (which calls `PrefixCache::match`) and `deallocate()` are both called within the same scheduler thread during `schedule_request()` - they never execute concurrently.

**Block manager selection**: When `enable_disagg_pd=false` and `enable_kvcache_store=false` (as in the reported issue), `BlockManagerImpl` is used, which is designed for single-threaded access. The concurrent version (`ConcurrentBlockManagerImpl`) with mutex protection is only used for disaggregated prefill/decode modes.

**The race condition you describe cannot occur** because:
1. A sequence's `deallocate()` is called only when it finishes
2. `allocate_shared()` for new requests happens in the same thread
3. There's no concurrent access to the same block's ref_count during the check-then-decrement

The underflow guard at line 73-85 is a defensive check, not an indication of expected concurrent access.

xllm/core/framework/block/block_manager_pool.cpp
```diff
@@ -90,10 +90,13 @@ void BlockManagerPool::deallocate(std::vector<Sequence*>& sequences) {
 
 void BlockManagerPool::deallocate(Sequence* sequence) {
   DCHECK(sequence != nullptr);
-  // add blocks to the prefix cache
   int32_t dp_rank = get_dp_rank(sequence);
-  cache(sequence);
+  // deallocate must be called before cache() to correctly update
+  // num_used_blocks_ cache() increases ref_count, which would cause
```
weizhehuang0827, line 95: If a sequence hold a block, its `ref_count` is 1. After adding to prefix cache, its `ref_count` is 2. `block.ref_count() <= 2` is feasible in `deallocate`. Your issue might not be here？

QwertyJack: You're right that after `cache()`, ref_count = 2 (sequence + cache), and `2 <= 2` should pass the check.

Looking at the code more carefully, I believe the issue manifests with **concurrent shared block usage**:

1. Block in cache only, ref_count = 1
2. Sequence A matches it via `allocate_shared()`: ref_count becomes 2, `num_used_blocks_++`
3. Sequence B matches same block (while A still running): ref_count = 3, check `3 <= 2` fails, `num_used_blocks_++` **NOT called**
4. Sequence A completes:
   - **Buggy order**: `cache(A)` first (no-op, already cached), ref_count = 3
   - `deallocate(A)`: `3 <= 2`? FALSE → `num_used_blocks_--` **NOT called**
   - `reset(A)`: ref_count = 2
5. Sequence B completes:
   - `cache(B)`: no-op, ref_count = 2
   - `deallocate(B)`: `2 <= 2`? TRUE → `num_used_blocks_--`
   - `reset(B)`: ref_count = 1

Net effect: +1 (step 2) - 0 (step 4, skipped) - 1 (step 5) = 0. Still balanced!

Hmm, I'm having trouble finding the exact leak path. However, the user observed `kv_cache_utilization = 135%` with no running requests, which indicates `num_used_blocks_` is not properly decremented.

With the fix (deallocate before cache), ref_count at deallocation time is guaranteed to reflect only the sequence's own reference, making the accounting more predictable.

Would you be able to share more details about the test scenario or help identify the exact leak path?

QwertyJack:
Seems related to PR #827. Unable to reproduce on main. Close for now.

xllm/core/framework/block/block_manager_pool.cpp
yq33victor, line 95: This change appears incorrect, `num_used_blocks_` will have an incorrect value after the modification.

PR #839, opened 2026-01-30
refactor: optimize vlm_master interface.

xiao-yu-chen:


xllm/api_service/call.h
```diff
@@ -29,7 +29,7 @@ class Call {
   std::string get_x_request_id() { return x_request_id_; }
   std::string get_x_request_time() { return x_request_time_; }
 
-  std::string& get_request_payload() { return request_payload_; }
+  std::string take_request_payload() { return std::move(request_payload_); }
```
yq33victor, line 32: is seems no performance improve?

xiao-yu-chen: The modification here have nothing to do with performance. They are mainly made to fit the semantics of move

wly-115 (approved):

PR #862, opened 2026-02-02
bugfix: fix early return in prefetch_from_storage.

shifengmin:


yq33victor (approved):
ok, looks good.

BTW: Currently, there are actually no multiple sequences in a single request.

PR #869, opened 2026-02-03
feat: llama3-8b run through based on ilu device.

laneeeee:


xllm/models/llm/llama.h
```diff
@@ -0,0 +1,76 @@
+/* Copyright 2025 The xLLM Authors. All Rights Reserved.
```
XuZhang99, line 1: 2026

PR #876, opened 2026-02-03
feat: support mm_data rpc transport.

wly-115:


xllm/proto/worker.proto
```diff
@@ -218,10 +257,11 @@ message ForwardInput {
...
   uint64 batch_id = 38;
   int32 batch_forward_type = 39;
-  MMData mm_data = 40;
-  repeated int32 q_cu_seq_lens = 41;
-  repeated int32 dp_is_decode = 42;
-  repeated int32 kv_cache_tokens_nums = 43;
+  repeated int32 q_cu_seq_lens = 40;
+  repeated int32 dp_is_decode = 41;
+  repeated int32 kv_cache_tokens_nums = 42;
+  MMBatchData mm_data = 43;
```
DragonFive: Is it realy needed to change all these 4 fileds number? because the proto is also in rec predictor framework and stress test script, they all need change.

yq33victor (approved):
LGTM

PR #880, opened 2026-02-04
bugfix: fix request failure for CUDA backend when enable_graph is true.

Wang-1F:


xllm/core/runtime/cuda_graph_executor_impl.cpp
```diff
@@ -340,9 +344,8 @@ std::optional<ModelInputParams> CudaGraphPersistentParam::update(
         static_cast<int32_t>(block_size),  // block_size
         sliding_window,                    // window_size_left
         /*enable_cuda_graph=*/true,
-        causal,                     // causal
-        /*use_tensor_core=*/true);  // let use_tensor_core=true temporarily
-                                    // until we refactor flashinfer API
+        causal,                                // causal
+        /*use_tensor_core=*/use_tensor_core);  // use_tensor_core
```
XuZhang99: remove `/*use_tensor_core=*/` and `// use_tensor_core`

Wang-1F: Updated.

PR #883, opened 2026-02-04
feat: use CustomPagedAttention in ACL graph decode to avoid kv_seq_lens on CPU[5/N].

yingxudeng:


xllm/core/layers/common/attention_metadata.h
```diff
@@ -104,6 +104,11 @@ struct AttentionMetadata {
   // for npu
   torch::Tensor attn_mask;
   torch::Tensor kv_seq_lens_host;
+  // For ACL graph execution - tiling data for CustomPagedAttention
+  // When enable_acl_graph is true, this tensor is used instead of
+  // kv_seq_lens_host to avoid .to(kCPU) operations that break ACL graph capture
+  torch::Tensor paged_attention_tiling_data;
+  bool enable_acl_graph = false;
```
yq33victor: Is it enough to use `paged_attention_tiling_data.defined()` to make the judgment ?

xllm/core/layers/common/attention_metadata.h
```diff
@@ -104,6 +104,11 @@ struct AttentionMetadata {
   // for npu
```
zhang-minchao, line 112: add `#if defined(USE_NPU)` for npu

PR #895, opened 2026-02-07
feat: support index cache transfer in PD separate deployment scenario.

sunbaosong:


xllm/core/framework/kv_cache/llm_data_dist_transfer.cpp
```diff
@@ -79,6 +79,7 @@ void LlmDataDistTransfer::allocate_kv_cache(
     const int64_t num_layers,
     const std::vector<std::vector<int64_t>>& kv_cache_shape,
     torch::ScalarType dtype) {
+  enable_lighting_indexer_ = kv_cache_shape.size() > 2;
```
XuZhang99: it's not good to set `enable_lighting_indexer_`, why not pass `enable_lighting_indexe` in construct func?

PR #906, opened 2026-02-10
feat: optimized rejected sampler for Eagle3.

RobbieLeung:


XuZhang99 (approved):

xllm/core/framework/sampling/sampler.cpp
```diff
@@ -72,7 +72,7 @@ SampleOutput Sampler::forward(torch::Tensor& logits,
     auto greedy = greedy_sample(probs);
     samples = torch::where(params.do_sample, random, greedy);
   }
-  output.probs = probs;
+  output.probs = probs.to(logits.dtype());
```
Clement-Wang26, line 75: probs is float32 and logits is bf16, so should we cast logits to float32 instead?

Clement-Wang26 (approved):

PR #926, opened 2026-02-12
refactor: extract MTPWorkerImpl from SpeculativeWorkerImpl.

Clement-Wang26:
details:
- SpeculativeWorkerImpl is now a base class with common logic.
- MTPWorkerImpl handles MTP-specific draft model + validation.
- Eagle3WorkerImpl inherits from MTPWorkerImpl.
- worker.cpp uses factory pattern to dispatch by FLAGS_speculative_algorithm.

### DIFF
--- a/xllm/core/runtime/eagle3_worker_impl.cpp
+++ b/xllm/core/runtime/eagle3_worker_impl.cpp
@@ -46,18 +46,17 @@ runtime::Options eagle3_draft_options(const runtime::Options& options) {
 Eagle3WorkerImpl::Eagle3WorkerImpl(const ParallelArgs& parallel_args,
                                    const torch::Device& device,
                                    const runtime::Options& options)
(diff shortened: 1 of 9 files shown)


xllm/core/runtime/eagle3_worker_impl.cpp
```diff
@@ -1,4 +1,4 @@
-/* Copyright 2025 The xLLM Authors. All Rights Reserved.
+/* Copyright 2026 The xLLM Authors. All Rights Reserved.
```
XuZhang99: why change this?

Clement-Wang26: done.

yq33victor (approved):
LGTM!

RobbieLeung (approved):

PR #927, opened 2026-02-13
feat: support glm-5 bf16 for mlu.

a120092009:


phantomlei3:
will create a new PR for this support

PR #928, opened 2026-02-14
feat: support GLM5 on NPU with PD disaggregation, mtp, and reasoning parser.

sunbaosong:


yq33victor (approved):
LGTM

xllm/core/framework/kv_cache/llm_data_dist_transfer.cpp
```diff
@@ -53,7 +53,11 @@ LlmDataDistTransfer::LlmDataDistTransfer(const std::string& device_ip,
     LOG(INFO) << "Create LlmDataDistTransfer for mix instance.";
     role = LlmRole::kMix;
   }
-
+  // Note: we currently support glm5_moe for model with light indexer in
+  // LlmDataDistTransfer.
+  if (model_type_ == "glm5_moe") {
```
JimHsiung: The logic here for determining whether to enable enable_lighting_indexer is inconsistent with that in the upper-layer LLMEngine. It is recommended to pass this parameter directly from the upper layer to align the judgment logic, and to support all models containing lighting_indexer instead of only GLM5. https://github.com/jd-opensource/xllm/blob/main/xllm/core/distributed_runtime/llm_engine.cpp#L330

PR #931, opened 2026-02-24
feat: validate flags before start xllm server.

yq33victor:


xllm/xllm.cpp
```diff
@@ -107,6 +107,19 @@ std::string get_model_backend(const std::filesystem::path& model_path) {
   return ModelRegistry::get_model_backend(model_type);
 }
 
+void validate_flags(const std::filesystem::path& model_path) {
+  if (FLAGS_backend.empty()) {
+    LOG(FATAL) << "Model is not supported currently. "
+               << get_model_type(model_path);
+  }
+#if defined(USE_MLU)
+  if (FLAGS_block_size != 16) {
```
phantomlei3: FLAGS_block_size can also equal to 1 in MLU

phantomlei3:
LGTM

PR #932, opened 2026-02-24
bugfix: fix REC multi-round CUDA graph accuracy crash.

LMX-xin:
This PR fixes an accuracy mismatch in REC multi-round decode when enable_graph=true.

  Root cause:

  - decoder_reshape_and_cache assumed block_table is contiguous and read block_id with a stride-1 access pattern.
  - In CUDA graph replay, block_table may come from a sliced persistent tensor with non-unit stride.
  - This caused incorrect block_id reads, wrong KV cache writes, and decode divergence.

  Fix:

  - Update decoder_reshape_and_cache to read block_table using its actual stride (stride(0)), not a contiguous assumption.
  - Keep the kernel path compatible with both contiguous and non-contiguous block_table layouts.


DragonFive (approved):
LGTM

Wang-1F:
LGTM

XuZhang99 (approved):

zhang-minchao (approved):

PR #936, opened 2026-02-25
feat: add kv cache quantization to mlu.

a120092009:


xllm/core/runtime/worker_impl.cpp
```diff
@@ -89,54 +89,105 @@ bool WorkerImpl::allocate_kv_cache(
...
   CHECK(kv_caches_.empty()) << "KV caches are already initialized.";
 
+  // Check if KV cache quantization is enabled
+  // "auto" (default): cache dtype aligns with model dtype (no quantization)
+  // "int8": enables INT8 quantization
+  const bool enable_kv_cache_quant = options_.kv_cache_dtype() == "int8";
+
+  if (enable_kv_cache_quant) {
+#if !defined(USE_MLU)
+    CHECK(false) << "KV Cache quantization is only supported on MLU backend. "
```
XuZhang99: use `LOG(FATAL)`.

xllm/core/common/global_flags.cpp
```diff
@@ -561,4 +561,13 @@ DEFINE_int32(beam_width, 1, "Beam width for beam search.");
...
 DEFINE_int32(health_check_interval_ms,
              3000,
-             "Worker health check interval in milliseconds.");
\ No newline at end of file
+             "Worker health check interval in milliseconds.");
+
+// --- kv cache quantization config ---
+
+DEFINE_string(
+    kv_cache_dtype,
```
XuZhang99: add this flag below this line: https://github.com/jd-opensource/xllm/blob/b05b5f12fdd048a89be4a4f422f7738fa8f32630/xllm/core/common/global_flags.cpp#L128
and add `kCacheOptions` for these kvcache config flags.

xllm/core/runtime/worker_impl.cpp
phantomlei3: fixed

xllm/core/common/global_flags.cpp
phantomlei3: fixed

PR #949, opened 2026-02-27
feat: allow normal rope for deepseek models in mlu.

phantomlei3:
extract the rope fix from #927

### DIFF
--- a/xllm/core/layers/common/tests/qwen2_vision_attention_tests.cpp
+++ b/xllm/core/layers/common/tests/qwen2_vision_attention_tests.cpp
@@ -90,7 +90,10 @@ TEST_F(Qwen2VisionAttentionTest, ForwardTest) {
   int32_t num_tokens = batch_size * seq_len;
 
   auto hidden_states =
-      torch::randn({num_tokens, mm_hidden_size}, options_) * 0.02f;
+      test::seeded_tensor("qwen2_vision_attention.hidden_states",
+                          {num_tokens, mm_hidden_size},
+                          torch::kBFloat16,
+                          options_.device());
 
   // Create cu_seq_len (cumulative sequence lengths)
   std::vector<int32_t> cu_seq_len_vec = {0, seq_len, num_tokens};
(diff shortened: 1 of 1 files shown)


yq33victor (approved):
LGTM

xllm/core/layers/common/rotary_embedding.cpp
```diff
@@ -86,6 +86,40 @@ void RotaryEmbeddingImpl::forward(torch::Tensor& q,
...
+                                  int64_t max_query_len,
+                                  bool is_prompt) {
+  bool discrete;
+  std::optional<torch::Tensor> position_ids;
+  if (is_prompt) {
+    discrete = false;
+    if (Device::type_str() == "cuda" || Device::type_str() == "npu" ||
+        Device::type_str() == "ilu") {
+      position_ids = positions;
+    }
```
RobbieLeung, line 102: Is `position_ids` useless at MLU?

RobbieLeung (approved):

xllm/core/layers/common/rotary_embedding.cpp
phantomlei3: currently it is useless in prefill stage for MLU, but we have a refactored version of rope kernel which will be released soon. Then we will align with other backend which requires positions for prefill stage

phantomlei3:
```
    ┌─────────────────┬───────────┬────────────┬──────────────┬──────────┬──────────────────────────┐
    │ 阶段            │ is_prompt │ is_chunked │ position_ids │ discrete │ 位置计算方式             │
    ├─────────────────┼───────────┼────────────┼──────────────┼──────────┼──────────────────────────┤
    │ 标准 Prefill    │ True      │ False      │ None         │ False    │ 通过 cu_seqlens 隐式计算 │
    │ Chunked Prefill │ True      │ True       │ positions    │ True     │ 显式 positions           │
    │ Decode          │ False     │ -          │ positions    │ True     │ 显式 positions           │
    └─────────────────┴───────────┴────────────┴──────────────┴──────────┴──────────────────────────┘

```

PR #965, opened 2026-03-01
refactor: split prefill and decode attention on cuda device.

XuZhang99:


xllm/core/layers/cuda/flashinfer_planinfo.cpp
```diff
@@ -83,170 +79,194 @@ void update_plan_info(std::shared_ptr<PlanInfo> plan_info,
...
                               head_dim_qk,
                               head_dim_vo,
                               /*pos_encoding_mode=*/0,
                               /*use_sliding_window=*/false,
                               /*use_logits_soft_cap=*/false,
                               /*use_fp16_qk_reduction=*/false);
+    const int64_t batch_size = attn_meta.paged_kv_last_page_len.size(0);
+    torch::Tensor qo_indptr_host =
+        get_cache_buffer(batch_size + 1, torch::kCPU);
+    torch::Tensor qo_indptr = qo_indptr_host.to(torch::kCUDA);
```
zhang-minchao: 需要qo_indptr吗？看后面没有使用

XuZhang99: this line is useless in fact, but the original code has this line:
https://github.com/jd-opensource/xllm/blob/0b227312a41e2778a18cac89f1ecb717cfe09a62/xllm/core/layers/cuda/flashinfer_planinfo.cpp#L184
I will delete this line.

RobbieLeung (approved):

PR #966, opened 2026-03-01
feat: support chunked prefill on cuda device.

XuZhang99:


xllm/core/runtime/cuda_graph_executor_impl.cpp
```diff
@@ -466,10 +465,10 @@ std::optional<ModelInputParams> CudaGraphPersistentParam::update(
         << ", causal=" << causal << ", backend=" << backend
         << ", enable_cuda_graph=" << attn_metadata->enable_cuda_graph;
 
-    if (causal) {
+    if (attn_metadata->is_prefill) {
       layer::flashinfer::update_prefill_plan_info(
           attn_metadata->plan_info,
-          backend,
+          /*backend=*/"fa2",  // flashinfer paged fa3 is slow, use fa2 instead
```
zhang-minchao: So, the backend now defaults to `fa2` whenever `is_prefill` or `is_chunked_prefill` is true? Previously, the logic triggered `fa2` when causal was set to false, which I assume was the path for decode.

XuZhang99: No, it's just a mistake and I will correct it.

PR #983, opened 2026-03-03
feat: implement moe fused topk layer.

XuZhang99:


xllm/core/kernels/ops_api.cpp
```diff
@@ -412,6 +414,12 @@ std::tuple<torch::Tensor, torch::Tensor> moe_active_topk(
                               params.scoring_func,
                               params.route_scale,
                               params.e_score_correction_bias);
+#elif defined(USE_CUDA) || defined(USE_MUSA)
```
yq33victor, line 414: USE_MUSA ?

XuZhang99: musa device can use flashinfer ops directly.

RobbieLeung (approved):

PR #999, opened 2026-03-05
bugfix: add group_id when creating the HccLProcessGroup to support multiple communication process_group.

shan-chen-feng:
…ltiple communication process_group.

### DIFF
--- a/xllm/core/framework/parallel_state/npu_process_group.h
+++ b/xllm/core/framework/parallel_state/npu_process_group.h
@@ -50,6 +50,14 @@ class ProcessGroupImpl : public ProcessGroup {
                  std::vector<torch::Tensor>& outputs) override;
 
  private:
+  // group_id is a optional value of c10d_npu::ProcessGroupHCCL::Options,
+  // which is used to construct the key of process_domain, if the value
+  // was not set when initializing more than one HcclProcessGroup, A Hccl ERROR
+  // that inidicates the comm name has already exist would occur,
+  // so for each time when we create a new process group, we should increase
+  // group_id_ and pass it to c10d_npu::ProcessGroupHCCL::Options
+  static int32_t group_id_;
+
   HcclComm comm_ = nullptr;
   c10_npu::NPUStream comm_stream_;
 };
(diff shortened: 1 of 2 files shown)


xllm/core/framework/parallel_state/npu_process_group.h
```diff
@@ -50,6 +50,14 @@ class ProcessGroupImpl : public ProcessGroup {
                  std::vector<torch::Tensor>& outputs) override;
 
  private:
+  // group_id is a optional value of c10d_npu::ProcessGroupHCCL::Options,
+  // which is used to construct the key of process_domain, if the value
+  // was not set when initializing more than one HcclProcessGroup, A Hccl ERRROR
+  // that inidicates the comm name has already exist would occur,
+  // so for each time when we create a new process group, we should increase
+  // group_id_ and pass it to c10d_npu::ProcessGroupHCCL::Options
+  static int32_t group_id_;
```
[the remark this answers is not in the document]
shan-chen-feng: it is initialized in single process, no need to use atomic

xllm/core/framework/parallel_state/npu_process_group.h
```diff
@@ -50,6 +50,14 @@ class ProcessGroupImpl : public ProcessGroup {
                  std::vector<torch::Tensor>& outputs) override;
 
  private:
+  // group_id is a optional value of c10d_npu::ProcessGroupHCCL::Options,
+  // which is used to construct the key of process_domain, if the value
+  // was not set when initializing more than one HcclProcessGroup, A Hccl ERROR
+  // that inidicates the comm name has already exist would occur,
+  // so for each time when we create a new process group, we should increase
+  // group_id_ and pass it to c10d_npu::ProcessGroupHCCL::Options
+  static int32_t group_id_;
```
yq33victor, line 59: don't understand the modification. does it means we can not create multiple `ProcessGroupHCCL` ? How did  it work before.

shan-chen-feng: there won’t be explicit error when creating multiple process groups，but when using a second process group with any of the communication operation，the error occurs。 this error was not triggered because only tp process group was used for qwen3next

shan-chen-feng:
this bug was also fixed in https://github.com/jd-opensource/xllm/pull/891, but no body cares right

PR #1010, opened 2026-03-06
bugfix: improve qwen25 tool-call parsing robustness.

yingxudeng:


xllm/api_service/utils.h
walsonyang, line 138: error handling for result？

PR #1027, opened 2026-03-10
feat: use ixformer linear for ilu device.

laneeeee:


xllm/core/kernels/ilu/matmul.cpp
```diff
@@ -17,11 +17,57 @@ limitations under the License.
...
+  bool persistent = false;
+  std::vector<int64_t> output_shape = a.sizes().vec();
+  if (!output_shape.empty()) {
+    output_shape[output_shape.size() - 1] = b.size(0);
+  }
+  torch::Tensor output = a.new_empty(output_shape);
+
+  bool use_gemv = true;
+  const int64_t gemv_max_batch = 1;
+  const char* disable_infer_gemm_ex_str = std::getenv("DISABLE_INFER_GEMM_EX");
```
XuZhang99: use `get_string_env`:
https://github.com/jd-opensource/xllm/blob/7e2c710db42fa2f616ae94cff0f845eecf165082/xllm/core/util/env_var.cpp#L57

but I think `DISABLE_INFER_GEMM_EX` should be a bool var.

xllm/core/kernels/ilu/ixformer.h
```diff
@@ -60,6 +60,18 @@ torch::Tensor xllm_paged_attention(
...
+                              torch::Tensor& weight,
+                              int64_t act_type,
+                              const std::optional<torch::Tensor>& bias,
+                              const std::optional<torch::Tensor>& out,
+                              const std::optional<bool> persistent);
+
+torch::Tensor ixformer_linear_ex(torch::Tensor& input,
+                                 torch::Tensor& weight,
+                                 const c10::optional<torch::Tensor>& bias,
+                                 const c10::optional<torch::Tensor>& out);
```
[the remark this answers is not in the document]
laneeeee: it is iluvatar kernel's header file, can not modify parameter type

xllm/core/kernels/ilu/matmul.cpp
laneeeee: fixed it, thanks for your suggestion, but the var will be used in other llm inference framework, like vllm.

PR #1034, opened 2026-03-11
refactor: remove the outdated glm5 model type.

yq33victor:


xllm/xllm.cpp
```diff
@@ -54,7 +54,7 @@ static std::unordered_set<std::string> deepseek_like_model_set = {
     "deepseek_v32_mtp",
     "kimi_k2",
     "glm4_moe_lite",
-    "glm5_moe",
+    // glm5 model type
     "glm_moe_dsa",
```
XuZhang99: nit: add `// glm5 model type` at the end of this line.

PR #1038, opened 2026-03-11
feat: add unified request stats logs for prefill and service.

DongheJin:


xllm/core/scheduler/disagg_pd_scheduler.cpp
```diff
@@ -512,6 +512,9 @@ void DisaggPDScheduler::prefill_send_first_generation() {
     auto request = running_requests_[i];
     // Check if the request is a recently completed prefill request
     if (request->sequences()[0]->num_generated_tokens() == 1) {
+      if (request->state().stream) {
+        request->log_statistic(request->elapsed_seconds());
```
weizhehuang0827: If we need to print logs in the prefill instance, I think we can directly call `request->log_statistic` here regardless of whether it's a `stream`  request, and then remove the changes in `async_response_processor.cpp`.

xllm/core/scheduler/async_response_processor.cpp
```diff
@@ -117,8 +117,8 @@ void AsyncResponseProcessor::batch_process_completed_requests(
       HISTOGRAM_OBSERVE(
           end_2_end_latency_milliseconds,
           static_cast<int64_t>(end_2_end_latency_seconds * 1000.0));
-      if (request->finished() || request->cancelled()) {
-        // not log in prefill instance
+      if (request->finished() || request->cancelled() ||
+          request->sequences()[0]->num_generated_tokens() == 1) {
```
walsonyang: In the non-PD separation mode, is the logic still correct?

xllm/proto/chat.proto
```diff
@@ -147,6 +147,10 @@ message ChatRequest {
 
   // xllm_service rpc addr(ip:rpc_port) that forwards this request.
   optional string source_xservice_addr = 42;
+
+  // request trace headers from client.
+  optional string x_request_id = 43;
+  optional string x_request_time = 44;
```
walsonyang, line 153: There is another way to get x_request_id, http_request().GetHeader("x-request-id"). What are the differences between these two methods?

xllm/core/scheduler/disagg_pd_scheduler.cpp
DongheJin: done

xllm/core/scheduler/async_response_processor.cpp
DongheJin: fixed

xllm/proto/chat.proto
DongheJin: For xllm-service -> xllm request forwarding, useChatRequest from chat.proto.

JimHsiung (approved):

weizhehuang0827 (approved):

PR #1043, opened 2026-03-12
feat: support debug logs and dynamic switching.

magicheng0816:


xllm/server/xllm_server.cpp
```diff
@@ -50,7 +50,8 @@ bool XllmServer::start(std::unique_ptr<APIService> service) {
                             "sleep => SleepHttp,"
                             "wakeup => WakeupHttp,"
                             "link_d2d => LinkD2DHttp,"
-                            "unlink_d2d => UnlinkD2DHttp") != 0) {
+                            "unlink_d2d => UnlinkD2DHttp,"
+                            "logging => LoggingHttp") != 0) {
```
walsonyang, line 54: I'm wondering if it's necessary to implement logging via http interface, what if we manually associate the logs on the server by using the request_id or log_id?

magicheng0816: This is a control and viewing interface for logging, which facilitates turning on the debug log switch and viewing more detailed logs after logging into a machine

xllm/api_service/api_service.cpp
```diff
@@ -1195,4 +1196,110 @@ void APIService::UnlinkD2DHttp(::google::protobuf::RpcController* controller,
...
+  }
+  if (upper == "FATAL") {
+    FLAGS_minloglevel = 3;
+    FLAGS_v = 0;
+    return true;
+  }
+  return false;
+}
+
+}  // namespace
```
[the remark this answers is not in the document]
magicheng0816: just only triggered during manual debugging

PR #1045, opened 2026-03-12
bugfix: fix decode instance infinite retry allocate.

magicheng0816:


xllm/core/scheduler/disagg_pd_scheduler.h
```diff
@@ -147,6 +147,12 @@ class DisaggPDScheduler : public ContinuousScheduler {
   moodycamel::BlockingConcurrentQueue<std::shared_ptr<Request>>
       prefill_request_queue_offline_;
 
+  // Max retries when D returns 404 (e.g. try_allocate failed), avoid infinite
+  // retry.
+  static constexpr int kAddNewRequestsMaxRetryOnReject = 3;
```
XuZhang99, line 152: `int` ==> `int32_t` or `int64_t`

PR #1046, opened 2026-03-12
bugfix: avoid decode instance leak if prefill instance prefill fail.

magicheng0816:


JimHsiung:
When a Prefill instance fails, an unlink cluster is triggered, which clears legacy requests, seemingly contradicting the current implementation logic. https://github.com/jd-opensource/xllm/blob/main/xllm/core/scheduler/disagg_pd_scheduler.cpp#L790

PR #1059, opened 2026-03-13
feat: support new interface v1/sample.

RobbieLeung:
This update introduces a new `/v1/sample` interface and its Python integration as one complete feature.

`/v1/sample` is designed for placeholder-targeted token sampling: given a prompt and a literal selector (for example, `<emb_0>`), the server finds all matched positions and returns structured top-k token probabilities for each match in stable order (`choices[i].index == sample_id`). It covers full server-side support (proto/RPC/route/service, multi-match batch processing, multi-output response assembly, validation/error handling, docs, and tests), plus Python-side usability (`LLM.sample()`, pybind exposure of sample-related fields/types, and runnable example) so clients can use the feature directly without manual logprobs alignment/parsing.

### DIFF
--- a/xllm/api_service/sample_service_impl.cpp
+++ b/xllm/api_service/sample_service_impl.cpp
@@ -1,0 +1,471 @@
+/* Copyright 2026 The xLLM Authors. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://github.com/jd-opensource/xllm/blob/main/LICENSE
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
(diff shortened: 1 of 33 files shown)


yq33victor (approved):
LGTM for trigger CI build.

xllm/api_service/sample_service_impl.cpp
```diff
@@ -0,0 +1,538 @@
+/* Copyright 2025 The xLLM Authors. All Rights Reserved.
```
XuZhang99: 2026

xllm/api_service/sample_service_impl.cpp
```diff
@@ -0,0 +1,538 @@
...
+#include "core/distributed_runtime/llm_master.h"
+#include "core/framework/request/request_output.h"
+#include "core/framework/request/request_params.h"
+#include "core/framework/request/sample_slot.h"
+#include "core/util/uuid.h"
+
+namespace xllm {
+namespace {
+thread_local ShortUUID short_uuid;
+constexpr char kSelectorMatchFinishReason[] = "selector_match";
```
XuZhang99: why not use `std::string`?

XuZhang99 (approved):

PR #1061, opened 2026-03-15
refactor: replace NULL with nullptr.

Dragonliu2018:
Replace NULL with nullptr following "Effective Modern C++" Item 8 guidelines.
> Item 8: Prefer nullptr to 0 and NULL.

### DIFF
--- a/xllm/c_api/internal/helper.cpp
+++ b/xllm/c_api/internal/helper.cpp
@@ -74,7 +74,7 @@ void shutdown_log() {
 void set_init_options(BackendType backend_type,
                       const XLLM_InitOptions* init_options,
                       XLLM_InitOptions* xllm_init_options) {
-  if (nullptr == init_options) {
+  if (init_options == nullptr) {
     if (backend_type == BackendType::LLM) {
       memcpy(xllm_init_options,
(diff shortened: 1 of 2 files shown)


xllm/c_api/internal/helper.cpp
```diff
@@ -74,7 +74,7 @@ void shutdown_log() {
 void set_init_options(BackendType backend_type,
                       const XLLM_InitOptions* init_options,
                       XLLM_InitOptions* xllm_init_options) {
-  if (NULL == init_options) {
+  if (nullptr == init_options) {
```
XuZhang99: maybe `if (init_options == nullptr)` is better?

xllm/core/util/spin_rw_lock.h
```diff
@@ -93,7 +93,7 @@ typedef struct easy_spinrwlock_t {
 static __inline__ int easy_spinrwlock_rdlock(easy_spinrwlock_t* lock) {
   int ret = EASY_OK;
 
-  if (NULL == lock) {
+  if (nullptr == lock) {
```
XuZhang99: ditto.

PR #1066, opened 2026-03-16
refactor: prefer const_iterators to iterators.

Dragonliu2018:
Replace iterators with const_iterators following "Effective Modern C++" Item 13 guidelines.

> Item 13: Prefer const_iterators to iterators.

### DIFF
--- a/xllm/core/runtime/params_utils.cpp
+++ b/xllm/core/runtime/params_utils.cpp
@@ -636,13 +636,13 @@ void forward_output_to_proto(const torch::Tensor& next_tokens,
           pb_token.set_empty(true);
         }
         pb_token.mutable_top_tokens()->Reserve(token.top_tokens.size());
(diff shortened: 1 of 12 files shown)


xllm/core/runtime/params_utils.cpp
```diff
@@ -636,13 +636,13 @@ void forward_output_to_proto(const torch::Tensor& next_tokens,
           pb_token.set_empty(true);
         }
         pb_token.mutable_top_tokens()->Reserve(token.top_tokens.size());
-        for (auto it = token.top_tokens.begin(); it != token.top_tokens.end();
+        for (auto it = token.top_tokens.cbegin(); it != token.top_tokens.cend();
```
XuZhang99, line 639: `xllm::Slice` has no `cbegin` and `cend`, unless you add `cbegin()` and `cend` methods for `Slice`.

PR #1123, opened 2026-03-27
feat: support configurable max/min pixels for vlm image processors.

wly-115:


xllm/core/common/multimodal_config.cpp
```diff
@@ -0,0 +1,30 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+==============================================================================*/
+
+#include "multimodal_config.h"
+
+namespace xllm {
+
+std::string MMProcessConfig::to_string() const {
```
XuZhang99: this func should be placed to `multimodal_config.h`

wly-115: done.

XuZhang99 (dismissed):

PR #1130, opened 2026-03-30
perf: optimize qwen3.5 hybrid linear cache flow[4/N].

yingxudeng:


xllm/core/framework/kv_cache/hybrid_cache_utils.h
```diff
@@ -0,0 +1,126 @@
...
+  const int64_t dst_block_index = std::min<int64_t>(
+      max_block_index, static_cast<int64_t>((total_tokens - 1) / block_size));
+  int64_t src_block_index = dst_block_index;
+  if (prev_tokens > 0) {
+    src_block_index = std::min<int64_t>(
+        max_block_index, static_cast<int64_t>((prev_tokens - 1) / block_size));
+  }
+
+  return {block_ids[src_block_index], block_ids[dst_block_index]};
+}
```
JC-ut0, line 124: why do we need prev_tokens and total_tokens to calculate the src_block_index and dst_block_index for linear cache. Is this method safe for multi batch scenario?

yingxudeng: This leaves room for extending prefix cache support later.

PR #1212, opened 2026-04-07
feat: support Qwen3.5-VL model on npu device[6/N].

yingxudeng:


wly-115:
`torch::Tensor deepstack_process(torch::Tensor hidden_states,
                                  torch::Tensor visual_pos_masks,
                                  const torch::Tensor& visual_embeds)`
I remember Qwen 3.5 already removed DeepStack.

PR #1243, opened 2026-04-09
refactor: simplify xllm server startup routing and lifecycle helpers.

liutongxuan:
Unify API route selection with a table-driven strategy, extract signal and generic server setup helpers, and remove duplicate start logic for worker-like services while fixing listen address/port assignment after successful server start.


yq33victor (approved):
LGTM

JimHsiung (approved):

PR #1246, opened 2026-04-10
feat: add mlu mooncake pd push support.

phantomlei3:
- add backend-agnostic layer synchronizer wiring for layer-wise push overlap
- enable mooncake push on mlu with descriptor-based memory registration
- enforce mlu pd push flag validation and reject unsupported pull mode

### DIFF
--- a/xllm/core/framework/kv_cache/mooncake_transfer_engine.cpp
+++ b/xllm/core/framework/kv_cache/mooncake_transfer_engine.cpp
@@ -303,15 +303,6 @@ std::string MooncakeTransferEngine::initialize() {
 
 bool MooncakeTransferEngine::register_memory(std::vector<void*> addrs,
                                              std::vector<size_t> lens,
-                                             int64_t size_per_block) {
-  std::vector<uint64_t> buf_bytes(addrs.size(),
-                                  static_cast<uint64_t>(size_per_block));
-  return register_memory(
-      std::move(addrs), std::move(lens), std::move(buf_bytes));
-}
-
-bool MooncakeTransferEngine::register_memory(std::vector<void*> addrs,
-                                             std::vector<size_t> lens,
                                              std::vector<uint64_t> buf_bytes) {
   if (addrs.size() != lens.size() || addrs.size() != buf_bytes.size()) {
     LOG(ERROR) << "register_memory input size mismatch, addrs=" << addrs.size()
--- a/xllm/core/framework/kv_cache/mooncake_transfer_engine.cpp
+++ b/xllm/core/framework/kv_cache/mooncake_transfer_engine.cpp
@@ -158,7 +158,6 @@ bool MooncakeTransferEngineCore::open_session(const uint64_t cluster_id,
     return true;
   }
 
-  bool remote_opened = false;
   if (cluster_id != 0) {
     proto::MooncakeTransferEngineService_Stub* stub =
         get_or_create_stub_locked(cluster_id);
@@ -167,27 +166,24 @@ bool MooncakeTransferEngineCore::open_session(const uint64_t cluster_id,
       return false;
     }
 
-    proto::SessionInfo session_info;
-    session_info.set_addr(addr_);
+    proto::SessionInfo request;
+    request.set_addr(addr_);
     proto::Status response;
     brpc::Controller cntl;
-    stub->OpenSession(&cntl, &session_info, &response, nullptr);
+    stub->OpenSession(&cntl, &request, &response, nullptr);
     if (cntl.Failed() || !response.ok()) {
       LOG(ERROR) << "OpenSession failed, " << cntl.ErrorText();
       return false;
     }
-    remote_opened = true;
 
     LOG(INFO) << "OpenSession RPC to " << remote_addr
               << ", local_addr=" << addr_;
+    return true;
   }
 
   Transport::SegmentHandle handle = engine_->openSegment(remote_addr);
   if (handle == static_cast<Transport::SegmentHandle>(-1)) {
     LOG(ERROR) << "Fail to connect to " << remote_addr;
-    if (remote_opened) {
-      close_remote_session(this, cluster_id);
-    }
     return false;
   }
 
@@ -209,6 +205,18 @@ bool MooncakeTransferEngineCore::close_session(const uint64_t cluster_id,
             << ", remote_addr=" << remote_addr;
 
   auto it = handles_.find(remote_addr);
+  if (cluster_id != 0) {
+    if (it != handles_.end()) {
+      it->second.ref_count--;
+      LOG(INFO) << "Decremented ref_count for " << remote_addr
+                << ", ref_count=" << it->second.ref_count;
+      if (it->second.ref_count > 0) {
+        return true;
+      }
+    }
+    return close_remote_session(this, cluster_id);
+  }
+
   if (it == handles_.end()) {
     return true;
   }
@@ -217,14 +225,8 @@ bool MooncakeTransferEngineCore::close_session(const uint64_t cluster_id,
   LOG(INFO) << "Decremented ref_count for " << remote_addr
             << ", ref_count=" << it->second.ref_count;
 
-  // Keep the shared session alive while other users still reference it.
   if (it->second.ref_count > 0) {
     return true;
-  }
-
-  bool ret = true;
-  if (cluster_id != 0) {
-    ret = close_remote_session(this, cluster_id);
   }
 
   SegmentHandle handle = it->second.handle;
@@ -235,7 +237,7 @@ bool MooncakeTransferEngineCore::close_session(const uint64_t cluster_id,
 
   LOG(INFO) << "Closed session for " << remote_addr;
 
-  return ret;
+  return true;
 }
 
 SegmentHandle MooncakeTransferEngineCore::get_handle(
--- a/xllm/core/framework/kv_cache/mooncake_transfer_engine.cpp
+++ b/xllm/core/framework/kv_cache/mooncake_transfer_engine.cpp
@@ -15,25 +15,72 @@ limitations under the License.
 
 #include "mooncake_transfer_engine.h"
 
-#include <acl/acl.h>
-#include <gflags/gflags.h>
 #include <glog/logging.h>
 
+#include <algorithm>
 #include <numeric>
 
-#include "common/global_flags.h"
 #include "util/net.h"
 
 namespace xllm {
+
+namespace {
+
+bool close_remote_session(MooncakeTransferEngineCore* core,
+                          uint64_t cluster_id) {
+  proto::MooncakeTransferEngineService_Stub* stub =
+      core->get_or_create_stub(cluster_id);
+  if (stub == nullptr) {
+    LOG(ERROR) << "create_rpc_channel failed for cluster_id=" << cluster_id;
+    return false;
+  }
+
+  proto::SessionInfo session_info;
+  session_info.set_addr(core->addr());
+  proto::Status response;
+  brpc::Controller cntl;
+  stub->CloseSession(&cntl, &session_info, &response, nullptr);
+  if (cntl.Failed() || !response.ok()) {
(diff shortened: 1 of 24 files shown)


xllm/core/platform/mlu/mlu_layer_synchronizer.cpp
```diff
@@ -0,0 +1,46 @@
...
+    : events_(), event_record_flags_(static_cast<size_t>(num_layers)) {
+  events_.reserve(static_cast<size_t>(num_layers));
+  for (int64_t i = 0; i < num_layers; ++i) {
+    events_.emplace_back(c10::DeviceType::PrivateUse1);
+  }
+}
+
+bool MLULayerSynchronizerImpl::synchronize_layer(int64_t layer_index) {
+  while (!event_record_flags_[layer_index].load(std::memory_order_acquire)) {
+  }
```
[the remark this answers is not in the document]
phantomlei3: fixed

xllm/core/framework/kv_cache/mooncake_transfer_engine.cpp
```diff
@@ -99,54 +273,88 @@ bool MooncakeTransferEngineCore::initialize(int16_t listen_port,
...
 
-  auto it = handles_.find(remote_addr);
-  if (it != handles_.end()) {
-    // Session exists, just increment ref count
-    it->second.ref_count++;
-    LOG(INFO) << "Reusing existing session for " << remote_addr
-              << ", ref_count=" << it->second.ref_count;
-    return true;
+  {
+    std::unique_lock<std::mutex> lock(handle_mu_);
```
Clement-Wang26: I wonder if this locking scheme is over-engineered for open_session. This looks like a control-plane, low-frequency path, and the amount of metadata exchanged here is tiny, so the RPC cost should also be very small. Given that, I’m not sure we need the extra complexity of split locking plus the opening state and cv coordination. Holding the mutex for the whole function may be a simpler tradeoff here and would make the code much easier to follow and maintain, unless there is a concrete requirement for concurrent session establishment to different remote_addrs.

xllm/core/framework/kv_cache/mooncake_transfer_engine.cpp
```diff
@@ -155,68 +363,104 @@ bool MooncakeTransferEngineCore::open_session(const uint64_t cluster_id,
...
-
-  // Decrement ref count
-  it->second.ref_count--;
-  LOG(INFO) << "Decremented ref_count for " << remote_addr
-            << ", ref_count=" << it->second.ref_count;
-
-  // Only close when ref_count reaches 0
-  if (it->second.ref_count > 0) {
-    return true;
+  SegmentHandle handle = static_cast<SegmentHandle>(-1);
```
Clement-Wang26: Same comment as open_session.

xllm/core/framework/kv_cache/mooncake_transfer_engine.cpp
```diff
@@ -155,68 +363,104 @@ bool MooncakeTransferEngineCore::open_session(const uint64_t cluster_id,
...
   LOG(INFO) << "Closed session for " << remote_addr;
 
-  return true;
+  return ret;
 }
 
 SegmentHandle MooncakeTransferEngineCore::get_handle(
     const std::string& remote_addr) {
-  std::lock_guard<std::mutex> lock(mutex_);
+  std::lock_guard<std::mutex> lock(handle_mu_);
```
Clement-Wang26: I don’t think handle_mu_ needs to be renamed.

xllm/core/framework/kv_cache/mooncake_transfer_engine.cpp
```diff
@@ -260,29 +504,79 @@ std::string MooncakeTransferEngine::initialize() {
...
+
+  if (!register_memory_raw(std::move(addrs), std::move(lens), 0)) {
+    return false;
+  }
+
+  std::unordered_map<int32_t, GroupDesc> groups_map;
+  for (const BufferDesc& buffer : buffers) {
+    GroupDesc& group = groups_map[buffer.group_id];
+    group.group_id = buffer.group_id;
+    group.buffers.emplace_back(buffer);
```
Clement-Wang26: why call group？

xllm/core/framework/kv_cache/mooncake_transfer_engine.cpp
```diff
@@ -394,59 +734,70 @@ bool MooncakeTransferEngine::move_memory_blocks(
...
-  }
-
   TransferRequest::OpCode opcode;
   if (move_opcode == MoveOpcode::WRITE) {
     opcode = TransferRequest::WRITE;
   } else {
     opcode = TransferRequest::READ;
   }
 
+  std::vector<int64_t> addr_ids;
```
Clement-Wang26: why move？

xllm/core/framework/kv_cache/mooncake_transfer_engine.cpp
```diff
@@ -619,7 +970,7 @@ bool MooncakeTransferEngine::push_memory_blocks(
 
 void MooncakeTransferEngineService::OpenSession(
```
Clement-Wang26, line 681: I think this compatibility check would be better done during `link` / session handshake instead of being deferred to the first KV transfer via `ensure_remote_kv_mem()`.

From the current initialization flow, KV cache registration already happens before link: we create the transfer, allocate KV cache, and register KV memory first, and only then call `link_cluster()`. Given that ordering, the layout information is already available at link time.

With the current design, `link` only means "session established", while layout incompatibilities (for example block-size-derived `bytes_per_block` mismatch) are only surfaced later on the first push/pull. That makes the failure late and a bit surprising for the scheduler, because an instance can appear successfully linked but still be unusable for KV transfer.

A cleaner contract would be:
1. D sends its registered memory info together with the remote address during link/open-session.
2. P validates the remote layout immediately against its local registered KV layout.
3. If validation fails, `link_instance()` fails directly.

This would make the semantics of `link` much stronger: once link succeeds, the peer is not only reachable but also KV-transfer-compatible. It would also avoid a delayed runtime failure path in `ensure_remote_kv_mem()` / `move_memory_blocks()`.

If dynamic re-registration is possible later, we can handle that separately via versioning / cache invalidation / forced relink, but for the static initialization path this feels like a more appropriate place for a one-time compatibility check.

xllm/proto/mooncake_transfer_engine.proto
```diff
@@ -7,11 +7,37 @@ option cc_generic_services = true;
...
+message OpenSessionRequest {
+  string transfer_addr = 1;
+  uint64 cluster_id = 2;
+}
+
+message CloseSessionRequest {
+  string transfer_addr = 1;
+}
+
+message GetRegisteredMemoryRequest {}
```
Clement-Wang26: remove？

xllm/proto/mooncake_transfer_engine.proto
```diff
@@ -7,11 +7,37 @@ option cc_generic_services = true;
 
 import "common.proto";
 
-message SessionInfo {
-  string addr = 1;
+message OpenSessionRequest {
+  string transfer_addr = 1;
+  uint64 cluster_id = 2;
```
Clement-Wang26: I think the remote KV layout compatibility check should be done during `OpenSession` / link handshake, instead of being deferred to `ensure_remote_kv_mem()` on the first transfer.

From the current flow, KV cache is already allocated and registered before `link_cluster()` is called, so the layout information should already be available at handshake time. Doing the validation there would give us a cleaner invariant: once link succeeds, the peer is not only reachable but also KV-transfer-compatible.

If we move this validation into `OpenSession`, this new session-level role of `cluster_id` may also become unnecessary. Right now it is stored and later queried mainly to support the deferred `remote_addr -> cluster_id -> GetRegisteredMemory` path. If `RegisteredMemoryInfo` is exchanged and validated during handshake, that extra lookup path can likely be removed as well.

xllm/xllm.cpp
```diff
@@ -80,6 +80,31 @@ void validate_flags(const std::string& model_type) {
     LOG(FATAL) << "Currently, block_size must be 16 for MLU backend, we will "
                   "support other block sizes in the future.";
   }
+  if (FLAGS_enable_disagg_pd) {
```
Clement-Wang26, line 122: A small suggestion: can we directly adjust the gflag settings to the supported configuration here, instead of requiring users to provide the exact combination themselves? That would help reduce user friction and avoid unnecessary usage cost.

xllm/core/platform/mlu/mlu_layer_synchronizer.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+#include "platform/mlu/mlu_layer_synchronizer.h"
+
+#include <framework/core/MLUStream.h>
+
+#include <thread>
+
+namespace xllm {
+
+MLULayerSynchronizerImpl::MLULayerSynchronizerImpl(int64_t num_layers)
```
Clement-Wang26, line 24: Can we abstract `MLULayerSynchronizerImpl` / `NPULayerSynchronizerImpl` into a common `LayerSynchronizerImpl`, and keep the backend-specific event handling inside each platform implementation? The upper-level logic seems to rely on a fairly small shared interface, so this may help reduce platform-specific branching here.

xllm/core/framework/kv_cache/mooncake_transfer_engine.cpp
phantomlei3: layer would be better.

xllm/core/framework/kv_cache/mooncake_transfer_engine.cpp
phantomlei3: will fixed

xllm/xllm.cpp
phantomlei3: will fix

xllm/proto/mooncake_transfer_engine.proto
phantomlei3: will fix

xllm/core/framework/kv_cache/mooncake_transfer_engine.cpp
phantomlei3: fixed

phantomlei3 (commented):
fixed for current reviews

xllm/core/platform/mlu/mlu_layer_synchronizer.cpp
phantomlei3: not fix after negotiation

xllm/core/framework/kv_cache/mooncake_transfer_engine.cpp
```diff
@@ -106,40 +151,43 @@ bool MooncakeTransferEngineCore::open_session(const uint64_t cluster_id,
...
+    proto::Status response;
     brpc::Controller cntl;
-    stub->OpenSession(&cntl, &proto_session_info, &status, nullptr);
-    if (cntl.Failed() || !status.ok()) {
+    stub->OpenSession(&cntl, &session_info, &response, nullptr);
+    if (cntl.Failed() || !response.ok()) {
       LOG(ERROR) << "OpenSession failed, " << cntl.ErrorText();
       return false;
     }
+    remote_opened = true;
```
Clement-Wang26, line 178: In fact，we do not need to let decode node open  memory segment of prefill node. `return true ` is enough.

phantomlei3: fixed

xllm/core/framework/kv_cache/mooncake_transfer_engine.h
```diff
@@ -105,27 +107,27 @@ class MooncakeTransferEngine final {
   bool register_memory(std::vector<void*> addrs,
```
Clement-Wang26: this function should delete？

phantomlei3: removed

Clement-Wang26 (approved):

JimHsiung (approved):

PR #1248, opened 2026-04-10
bugfix: remove spurious backslash breaking output redirection in launch scripts.

### LINKED ISSUE #1247: fix: Shell syntax error in launch_xllm.md - broken output redirection (\ > LOG)
## Bug Description

In `docs/en/getting_started/launch_xllm.md`, the launch scripts for **NPU**, **NVIDIA GPU**, and **MLU** all contain a shell syntax error on the final argument line of the `xllm` command:

```bash
--node_rank=$i \ > $LOG_FILE 2>&1 &
```

### Problem

The `\\ >` sequence (two characters: backslash + space + `>`) causes the shell to interpret `>` as a *literal escaped greater-than character* rather than a **redirection operator**. This means:

1. The log file is **never created**.
2. All xllm output (stdout + stderr) is printed to the terminal instead of being captured.
3. The `&` still backgrounds the process, but you lose all logs.

You can reproduce the breakage with a minimal example:
```bash
echo hello \ > /tmp/test.log   # broken: writes nothing to file, prints 'hello >' to stdout
echo hello > /tmp/test.log      # correct
```

### Affected locations (lines 46, 87, 123)

All three platform scripts (NPU, NVIDIA GPU, MLU) have the same bug.

### Fix

Remove the spurious `\` before `>`:

```bash
# Before (broken):
    --node_rank=$i \ > $LOG_FILE 2>&1 &

# After (correct):
    --node_rank=$i > $LOG_FILE 2>&1 &
```

This preserves the command-line argument for `--node_rank` while correctly redirecting stdout+stderr to the log file.

kuishou68:
Closes #1247

## Problem

In `docs/en/getting_started/launch_xllm.md`, all three platform launch scripts (NPU, NVIDIA GPU, MLU) contain a shell syntax error on the final argument line:

```bash
--node_rank=$i \ > $LOG_FILE 2>&1 &
```

The `\ >` sequence causes the shell to interpret `>` as a literal escaped character rather than a **redirection operator**. As a result:
- The log file is **never created**
- All xllm stdout/stderr prints to the terminal instead of being captured
- Users running the launch script lose all logs

## Fix

Remove the spurious `\` before `>` in all three script blocks (lines 46, 87, 123):

```bash
# Before (broken):
    --node_rank=$i \ > $LOG_FILE 2>&1 &

# After (correct):
    --node_rank=$i > $LOG_FILE 2>&1 &
```

## Testing

```bash
# Demonstrate the bug (no file created):
echo hello \ > /tmp/test_broken.log
cat /tmp/test_broken.log  # file empty or missing

# Correct form:
echo hello > /tmp/test_fixed.log
cat /tmp/test_fixed.log   # prints: hello
```

### DIFF
--- a/docs/en/getting_started/launch_xllm.md
+++ b/docs/en/getting_started/launch_xllm.md
@@ -43,7 +43,7 @@ do
     --enable_chunked_prefill=true \
     --enable_schedule_overlap=true \
     --enable_shm=true \
-    --node_rank=$i \ > $LOG_FILE 2>&1 &
+    --node_rank=$i > $LOG_FILE 2>&1 &
 done
 ```
 
@@ -84,7 +84,7 @@ do
     --enable_prefix_cache=false \
     --enable_chunked_prefill=false \
     --enable_schedule_overlap=true \
-    --node_rank=$i \ > $LOG_FILE 2>&1 &
+    --node_rank=$i > $LOG_FILE 2>&1 &
 done
 ```
 
@@ -120,6 +120,6 @@ do
     --nnodes=$NNODES \
     --master_node_addr=$MASTER_NODE_ADDR \
     --block_size=16 \
-    --node_rank=$i \ > $LOG_FILE 2>&1 &
+    --node_rank=$i > $LOG_FILE 2>&1 &
 done
 ```


kuishou68:
Hi! Just following up on this PR — it removes the spurious backslash that was breaking output redirection in the launch scripts. The automated review (Gemini) found no issues. Would appreciate a maintainer review when you get a chance. Thanks!

XuZhang99:
Hi @kuishou68, we couldn't reproduce this issue in our all docker image.

kuishou68:
Thanks for checking. The issue I observed was on a bare Ubuntu 22.04 host running the launch scripts directly (not via Docker image). The backslash on line N breaks the shell redirection when the script is run with `bash` outside a container, since `bash` interprets the backslash differently than the Docker entrypoint. If the Docker image wraps the script in a way that skips the problematic redirection path, it would explain why it's not reproducible there. Happy to add a test case or provide the exact environment if that helps narrow it down.

XuZhang99:
> Thanks for checking. The issue I observed was on a bare Ubuntu 22.04 host running the launch scripts directly (not via Docker image). The backslash on line N breaks the shell redirection when the script is run with `bash` outside a container, since `bash` interprets the backslash differently than the Docker entrypoint. If the Docker image wraps the script in a way that skips the problematic redirection path, it would explain why it's not reproducible there. Happy to add a test case or provide the exact environment if that helps narrow it down.

Currently, xllm only can run on docker images provided by us.

kuishou68:
Understood — if xllm is only supported inside the provided Docker images, then the bare-host script path isn't a supported use case and this fix isn't needed. Happy to close the PR.

PR #1254, opened 2026-04-10
feat: support Qwen down_proj fallback for compressed-tensors ignored modules.

yingxudeng:


DragonFive (approved):
LGTM

xllm/core/framework/hf_model_loader_test.cpp
```diff
@@ -42,6 +42,10 @@ TEST(HFModelLoaderTest, LoadCompressedTensorsFp8StaticConfig) {
             }
           }
         },
+        "ignore": [
```
walsonyang, line 45: Is this rule applicable to all models?

yingxudeng: This field is part of the quantization config schema, not a model-specific rule. The JSON (including `ignore`) is generated by the quantization tool — at least [AngelSlim](https://github.com/Tencent/AngelSlim) produces this field. So the applicability depends on which quant tool was used, not the model itself.

walsonyang (approved):

PR #1264, opened 2026-04-13
bugfix: fix mtp prefix cache prefill starvation.

phantomlei3:


xllm/core/scheduler/prefill_only_scheduler.cpp
```diff
@@ -63,8 +63,9 @@ void PrefillOnlyScheduler::handle_prefill_requests(
 
   while (!waiting_priority_queue.empty() && remaining_seq_budget > 0 &&
          remaining_token_budget > 0 && latency_budget > estimate_latency) {
-    if (kv_cache_manager_->kv_cache_utilization() >=
-        FLAGS_prefill_scheduling_memory_usage_threshold) {
+    if (!enable_prefix_cache_ &&
+        kv_cache_manager_->kv_cache_utilization() >=
+            FLAGS_prefill_scheduling_memory_usage_threshold) {
```
RobbieLeung: `!enable_prefix_cache_` should be removed, we just need to promise that `kv_cache_manager_->kv_cache_utilization()` is right.

xllm/core/framework/block/block_manager_impl.cpp
```diff
@@ -194,4 +166,49 @@ void BlockManagerImpl::free(int32_t block_id) {
...
+  for (size_t i = begin; i < blocks.size(); ++i) {
+    const auto& block = blocks[i];
+    if (!block.is_valid()) {
+      continue;
+    }
+
+    const int32_t block_id = block.id();
+    CHECK_GE(block_id, 0);
+    CHECK_LT(static_cast<size_t>(block_id), active_seq_refs_.size());
+    if (active_seq_refs_[block_id]++ == 0) {
```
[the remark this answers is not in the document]
phantomlei3: fixed

xllm/core/framework/block/block_manager_impl.cpp
```diff
@@ -194,4 +166,49 @@ void BlockManagerImpl::free(int32_t block_id) {
...
+    const int32_t block_id = block.id();
+    CHECK_GE(block_id, 0);
+    CHECK_LT(static_cast<size_t>(block_id), active_seq_refs_.size());
+    if (active_seq_refs_[block_id] == 0) {
+      LOG(ERROR) << "active_seq_refs_==0 cannot release id:" << block_id
+                 << ", total block size: " << num_total_blocks();
+      std::unordered_set<int32_t> block_id_set;
+      block_id_set.insert(block_id);
+      std::string error_msg = "Block already released: ";
+      for (auto& id : free_blocks_) {
```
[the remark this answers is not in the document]
phantomlei3: fixed

xllm/core/framework/block/block_manager_impl.cpp
```diff
@@ -194,4 +166,49 @@ void BlockManagerImpl::free(int32_t block_id) {
...
+        if (block_id_set.count(id) != 0) {
+          error_msg.append(std::to_string(id)).append(" ");
+        }
+      }
+      LOG(FATAL) << error_msg;
+    }
+    if (--active_seq_refs_[block_id] == 0) {
```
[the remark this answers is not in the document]
phantomlei3: fixed

xllm/core/scheduler/prefill_only_scheduler.cpp
phantomlei3: fixed

xllm/core/framework/block/block_manager_impl.cpp
```diff
@@ -61,35 +62,12 @@ std::vector<Block> BlockManagerImpl::allocate(size_t num_blocks) {
   }
 
   // const auto block_ids = allocate(num_blocks);
-  num_used_blocks_.fetch_add(num_blocks, std::memory_order_relaxed);
+  add_seq_refs(blocks);
```
yq33victor: what's this differs from the current implementation. 
Currently, when a block is assigned, the class's assignment operator automatically increments the reference count (ref count++). In theory, there's no difference compared to the current approach.
```
Block& Block::operator=(const Block& other) {
  if (this != &other) {
    dec_ref_count();

    id_ = other.id_;
    size_ = other.size_;
    manager_ = other.manager_;
    ref_count_ = other.ref_count_;

    memcpy(hash_value_, other.hash_value_, XXH3_128BITS_HASH_VALUE_LEN);

    inc_ref_count(); // <--------------------- here
  }
  return *this;
}
```

Under what circumstances would current implementation cause a problem?

phantomlei3: ```ref_count``` and ```active_seq_refs``` are counting different things.

  | Step | Holder of physical block B | ref_count(B) | Active sequences using B |
  |---|---|---:|---:|
  | 1 | Prefix cache only | 1 | 0 |
  | 2 | PrefixCache::match() returns a local shared_blocks vector | 2 | 0 |
  | 3 | sequence->add_kv_blocks(shared_blocks) copies B into the sequence state | 3 | 1 |
  | 4 | Later allocation fails, so we call deallocate(shared_blocks) first | 2 | 1 |
  | 5 | Then sequence->reset() releases the sequence-owned copy | 1 | 0 |

  The problem is that ```ref_count``` includes every Block handle copy:

  - the prefix-cache node
  - the temporary shared_blocks vector
  - the sequence-owned vector

  But ```num_used_blocks_``` is supposed to mean:
  "how many physical blocks are still occupied by live sequences"

  So in prefix-cache paths, ```ref_count``` is not a reliable proxy for sequence ownership. A temporary copy can increase```ref_count``` even though no extra sequence is using the block. That is why the old ```ref_count() <= 2``` heuristic can drift, while ```active_seq_refs_``` tracks the intended quantity directly.

yq33victor (dismissed):
LGTM

xllm/core/framework/block/block_manager_impl.cpp
```diff
@@ -187,6 +215,10 @@ Block BlockManagerImpl::allocate() {
 void BlockManagerImpl::free(int32_t block_id) {
   // do nothing for reserved block 0
   if (block_id != 0) {
+    if (clear_used(&usage_accounted_ids_, block_id)) {
+      CHECK_GT(num_used_blocks_.load(std::memory_order_relaxed), 0u);
+      num_used_blocks_.fetch_sub(1, std::memory_order_relaxed);
```
yq33victor, line 225: no need call `num_used_blocks_.fetch_sub(1, std::memory_order_relaxed);` here, we have already call this in func  `BlockManagerImpl::deallocate`
```
void BlockManagerImpl::deallocate(const Slice<Block>& blocks) {
  if (options_.enable_prefix_cache()) {
        .....
        num_used_blocks_.fetch_sub(1, std::memory_order_relaxed); // <------
      }
    }
  } else {
    num_used_blocks_.fetch_sub(blocks.size(), std::memory_order_relaxed); // <------
  }
```

call functions links as follows:
```
deallocate (call num_used_blocks_.fetch_sub) 
->  ~Block() 
-> Block::dec_ref_count() 
-> manager_->free ( current function, call num_free_blocks_.fetch_add)
```

phantomlei3: we have been talked about this on talks. It is resolved

RobbieLeung (approved):
LGTM

liutongxuan (approved):

PR #1266, opened 2026-04-13
feat: add startup profile run for mlu llm engine.

phantomlei3:


xllm/core/common/global_flags.cpp
```diff
@@ -251,6 +251,8 @@ DEFINE_bool(enable_profile_token_budget,
             false,
             "Whether to enable profile token budget.");
 
+DEFINE_bool(enable_profile_run, false, "Whether to enable profile run.");
```
XuZhang99, line 254: "Whether to enable profile run when start xllm server."

PR #1271, opened 2026-04-13
bugfix: add ILU process group ctor for explicit local rank. (#1240)

liutongxuan:
Add the ILU ProcessGroup constructor overload that accepts local_rank and explicit group_ranks, so the DiT process-group creation path matches the expected signature and builds successfully.


RobbieLeung (dismissed):
LGTM

RobbieLeung (approved):

JimHsiung (approved):

PR #1290, opened 2026-04-16
optimize: CP support dynamic EP.

shifengmin:
Dynamic EP can be enabled when CP SIZE equals EP SIZE and WORLD SIZE to reduce peak memory usage and improve prefill performance.


DongheJin (approved):

yq33victor (approved):
LGTM

PR #1298, opened 2026-04-16
bugfix: add calculation of mm_embedding size in shm manager.

shan-chen-feng:


xllm/core/runtime/embed_vlm_worker_impl.cpp
```diff
@@ -97,10 +96,12 @@ std::optional<ForwardOutput> EmbedVLMWorkerImpl::step(
         sample_output.mm_embeddings.emplace_back(image_embed);
         token_start_idx += seq_len;
       }
+      output.sample_output = sample_output;
+    } else {
+      sample_output.embeddings = embeddings;
+      output.sample_output = sample_output;
+      output.embedding = embeddings;
```
wly-115, line 103: Can we remove the output.embedding = embeddings ？

PR #1331, opened 2026-04-22
refactor: add model input param groups and factory.

liutongxuan:
Introduce LLM/VLM/DIT/REC model-input param group types plus a ModelInput wrapper and ModelInputFactory to build model-specific payload partitions from legacy ModelInputParams without modifying the legacy struct.


xllm/core/framework/model/model_input.h
XuZhang99, line 1: Is it better to move `model_input*` files to a new dir, like `framework/model_input`?

liutongxuan: yes, subsequently, a series of refactoring operations will be carried out. currently, only new related classes are added

PR #1336, opened 2026-04-23
feat: support torch_npu 2.9.0.

WiseEnjoy:


CMakeLists.txt
```diff
@@ -255,8 +247,10 @@ find_package(Boost REQUIRED)
 find_package(Boost REQUIRED COMPONENTS serialization)
 find_package(Threads REQUIRED)
 # find all dependencies from vcpkg
+find_package(folly CONFIG REQUIRED)
 find_package(glog CONFIG REQUIRED)
 find_package(gflags CONFIG REQUIRED)
+find_package(fmt CONFIG REQUIRED)
```
XuZhang99: why need `fmt`?

third_party/CMakeLists.txt
```diff
@@ -59,3 +59,4 @@ if(USE_ILU)
 endif()
 
 target_link_libraries(mooncake_store PUBLIC transfer_engine cachelib_memory_allocator)
+target_link_libraries(cachelib_memory_allocator PRIVATE Folly::folly glog::glog)
```
XuZhang99, line 66: what is `cachelib_memory_allocator`?

CMakeLists.txt
WiseEnjoy: folly depends fmt, this line is deleted for now.

third_party/CMakeLists.txt
WiseEnjoy: one lib of monocake, for xllm link problem add this line, see comment:

```
## -> The linker scans each library only once and extracts object files solely to resolve currently undefined symbols.
## -> In modern CMake, the best practice is to explicitly declare a target's dependencies, allowing CMake to automatically determine the correct linking order. 
## -> The following code is added to prevent symbol resolution errors when xllm links error against mooncake-related static libraries.
```

DongheJin (approved):

yq33victor (approved):
LGTM

PR #1377, opened 2026-04-29
feat: add qwen3.5 for mlu backend without GDN.

phantomlei3:


xllm/core/layers/mlu/qwen3_5_fused_moe.h
```diff
@@ -0,0 +1,47 @@
...
+  void load_state_dict(const StateDict& state_dict) override;
+
+ protected:
+  void final_comm_allreduce(torch::Tensor& final_hidden_states,
+                            const torch::Tensor& hidden_states,
+                            torch::Tensor& shared_expert_output) override;
+
+ private:
+  void load_experts(const StateDict& state_dict);
+  torch::nn::Linear shared_expert_gate_{nullptr};
```
XuZhang99, line 42: why not use `Linear`?

SPK-L: You mean why we use linear, or why not use the linear class that defined in linear.h？The shared_expert_gate does not need to be sharded for tensor parallelism.

RobbieLeung (approved):

XuZhang99 (approved):

PR #1431, opened 2026-05-12
refactor: align offline arguments with online gflags.

weizhehuang0827:


yq33victor (dismissed):
LGTM

xllm/pybind/args.py
```diff
@@ -1,29 +1,43 @@
 import argparse
 from argparse import Namespace
 
+
+def _str_to_bool(value):
```
XuZhang99: change to `def _str_to_bool(value: str | bool) -> bool:`

weizhehuang0827: done

XuZhang99 (approved):

yq33victor (approved):
lgtm

PR #1433, opened 2026-05-13
feat: qwen3.5 support chunked prefill.

maojunx99:


xllm/core/layers/npu_torch/attention.cpp
```diff
@@ -91,24 +146,44 @@ void AttentionImpl::prefill_forward(torch::Tensor& query,
...
+        std::nullopt,
+        cumulative_lengths(attn_metadata.q_seq_lens_vec),
+        cumulative_lengths(attn_metadata.kv_seq_lens_vec),
+        num_heads_,
+        num_kv_heads_,
+        scale_,
+        0,
+        3,
+        "TND");
+    output.copy_(std::get<0>(fia_result).view_as(output));
```
RobbieLeung, line 110: It looks like that `npu_fused_infer_attention` needs host `q_cu_seq_lens` and 'kv_cu_seq_lens'. 
1. Host tensors confict with acl graph.
2. Use `q_cu_seq_lens` directly or prepare `q_cu_seq_lens`  at the prepraration of attention metadata.

xllm/core/layers/npu_torch/attention.cpp
```diff
@@ -91,24 +146,44 @@ void AttentionImpl::prefill_forward(torch::Tensor& query,
...
-                                     value,
-                                     attn_metadata.attn_mask,
-                                     attn_metadata.kv_seq_lens_host,
-                                     scale_,
-                                     output);
+    auto fia_result = xllm::kernel::npu::npu_fused_infer_attention(
+        query,
+        key,
+        value,
+        get_fia_split_fuse_attn_mask(query),
```
RobbieLeung: Attention mask should be prepared at the start of model forward.

xllm/core/layers/npu_torch/attention.cpp
maojunx99: Moved the preparation of host vector q_cu_seq_lens to attention metadata builder.

xllm/core/layers/npu_torch/attention.cpp
maojunx99: Addressed. The attention mask is now prepared at the start of model forward.

PR #1445, opened 2026-05-14
feat: support the wan22's dit and tp parallel.

ethan686:
including three changes, dit transformer impl, tp parallel for dit model and all the model args changes for wan2.2.  Note:  use mindiesd's tensor as dit's input, no tp, 10 transformer layer,  the cos similarity is 100% match with mindiesd.


XuZhang99 (dismissed):

yq33victor (dismissed):
LGTM

XuZhang99 (approved):

yq33victor (approved):
LGTM

PR #1452, opened 2026-05-14
feat: support text_encoder for flux2

wang-shuibin:
Compared with the side of MindIE SD, it has been verified that the precision of the text encoder on the xLLM side reaches 99.99%.


wang-shuibin (commented):
This PR adapts text_encoder(Mistral3 model) for the Flux2 model, and completes the accuracy verification. Compared with the side of MindIE SD, it has been verified that the precision of the text encoder on the xLLM side reaches 99.99%.

PR #1463, opened 2026-05-15
feat: support Flux2 in other common components.

wang-shuibin:
This PR implements adaptation for Flux2 in common components. Meanwhile, accuracy verification for the Flux2 model has been finished Flux2 model.


wang-shuibin (commented):
This PR adapts common components for Flux2 and completes accuracy verification for the Flux2 model. Additionally, it fixes a bug in xllm/core/layers/common/add_matmul.cpp.

PR #1471, opened 2026-05-18
feat: CP support prefix cache, chunked prefill and KV cache sharding.

shifengmin:
---

## Context Parallelism with Prefix Cache and KV Cache Sharding

This PR extends xLLM's Context Parallelism (CP) implementation to support prefix cache reuse and flexible KV cache sharding in both standard and PD-disaggregated deployments.

### Key Features

**KV Cache Sharding (`kv_split_size`)**
CP ranks can now shard the KV cache across a configurable number of ranks instead of replicating it. When `kv_split_size > 1`, each rank stores only `1/kv_split_size` of the KV blocks, reducing peak HBM consumption proportionally. A prefix AllGather (`in_prefix_slots`) is inserted before attention to collect the required prefix blocks from sibling ranks.

**Prefix Cache / Chunked Prefill Compatibility**
`cp_input_partition` now correctly partitions token sequences and block tables per CP rank while preserving prefix cache hit blocks. `recompute_new_cache_slots` and `compute_in_prefix_slots` translate physical block addresses into per-rank slot indices so that prefix cache hits are usable across shards.

**PD-Disagg KV Push Scheduling (`BuildPushSchedule`)**
KV blocks are pushed from prefill to decode using a structured schedule that groups layers by destination worker. When `kv_split_size > 1`, sibling CP ranks would all push to the same decode workers simultaneously causing an incast. `FLAGS_kv_push_dst_rotate` eliminates this by rotating each rank's traversal starting point, turning an N-to-1 incast into a balanced N-to-N pattern.

---

### New Flags

| Flag | Type | Default | Description |
|------|------|---------|-------------|
| `--kv_split_size` | int32 | `0` | KV-cache shard width. `0` = follow `cp_size` (legacy); `1` = no KV split, each rank holds a full replica; `K` (where K divides `cp_size`) = KV sharded across K ranks. |
| `--kv_push_dst_rotate` | bool | `true` | Rotate the destination worker traversal order in `push_kv_blocks` per KV-split rank to spread incast load evenly across decode workers. Flip to `false` to restore legacy push order. |
| `--kv_push_timing_log` | bool | `false` | Emit per-step (layer, dst, src_blocks, elapsed_ms) and per-call total timing logs for `push_kv_blocks`. Off by default; enable for incast / push-latency debugging. |

### Typical Launch Examples

**CP without KV split** (each rank holds full KV, prefix AllGather skipped):
```bash
--cp_size=4 --kv_split_size=1
```

**CP with KV split** (KV sharded across all 4 CP ranks, prefix AllGather enabled):
```bash
--cp_size=4 --kv_split_size=4
```

**Enable push timing debug**:
```bash
--kv_push_timing_log=true
```


xllm/core/common/global_flags.cpp
```diff
@@ -220,6 +220,51 @@ DEFINE_int32(ep_size, 1, "Expert parallel size for MoE model.");
 
 DEFINE_int32(cp_size, 1, "Context parallel size for DSA attention.");
 
+DEFINE_int32(prefill_cp_size,
```
shifengmin: unused, delete it

xllm/core/common/global_flags.cpp
```diff
@@ -220,6 +220,51 @@ DEFINE_int32(ep_size, 1, "Expert parallel size for MoE model.");
...
+             0,
+             "KV-cache split width. 0 falls back to cp_size (legacy); 1 means "
+             "no KV split (each CP rank stores full KV, skips prefix "
+             "AllGather); other K (K divides cp_size) means KV is sharded "
+             "across K ranks while token-CP still uses cp_size.");
+
+// Mirror of FLAGS_kv_split_size for the prefill side in PD-disaggregated
+// setups; decode instances must know this to compute correct stride when
+// pulling KV blocks from prefill (analogous to FLAGS_prefill_cp_size).
+DEFINE_int32(prefill_kv_split_size,
```
shifengmin: refactor: change to etcd message

third_party/xllm_atb_layers
shifengmin, line 1: update commit if atb layers PR merged
https://gitcode.com/xLLM-AI/xllm_atb_layers/pull/24

yq33victor (approved):
LGTM

Clement-Wang26 (approved):

PR #1475, opened 2026-05-18
refactor: remove negative condition when choosing decode or prefill.

rauletorresc:
Improve readability by: 
- Removing the negative condition when choosing between decode or prefill steps. 
- Removing the `else` block when returning from inside the `if` block.


rauletorresc:
Hi @XuZhang99, do you think this PR is relevant?

RobbieLeung (approved):

XuZhang99 (approved):

PR #1532, opened 2026-05-22
bugfix: drive pd push transfer by cursor to avoid miss transfer.

phantomlei3:


phantomlei3:
[need unit test on npu]

phantomlei3:
PR 1532 的重新编译已通过

xllm/core/framework/batch/batch_input_builder.cpp
```diff
@@ -207,48 +207,56 @@ BatchInputBuilder::BatchInputBuilder(
...
-  const size_t map_begin = std::max(win_begin, shared_blocks);
-  const size_t map_limit = std::min(local_size, shared_blocks + remote_size);
-  const size_t map_end = std::min(win_end, map_limit);
-
-  if (map_begin >= map_end) {
+  const size_t map_end = std::min(win_end, local_size);
+  const size_t remote_stride =
+      static_cast<size_t>(util::kv_split_stride_for_kv_transfer());
+  const size_t remote_end = map_end * remote_stride;
+  CHECK_GE(remote_size, remote_end)
```
shifengmin, line 237: should be: CHECK_GE(**remote_end**, **remote_size**) since **remote_size** denotes the number of blocks needed by the remote, while **remote_end** denotes the number of blocks held locally.

xllm/core/framework/batch/batch_input_builder.cpp
```diff
@@ -207,48 +207,56 @@ BatchInputBuilder::BatchInputBuilder(
...
-    const size_t remote_idx = local_idx - shared_blocks;
+  info.remote_blocks_ids.reserve(block_cnt * remote_stride);
+  remote_idxs.reserve(block_cnt * remote_stride);
+  for (size_t local_idx = win_begin; local_idx < map_end; ++local_idx) {
     info.local_blocks_ids.emplace_back(local_block_ids[local_idx]);
-    info.remote_blocks_ids.emplace_back(
-        full_info.remote_blocks_ids[remote_idx]);
-    remote_idxs.emplace_back(remote_idx);
+    for (size_t offset = 0; offset < remote_stride; ++offset) {
+      const size_t remote_idx = local_idx * remote_stride + offset;
```
shifengmin, line 257: break if remote_idx >= full_info.remote_blocks_ids.size()

xllm/core/framework/batch/batch_input_builder.cpp
phantomlei3: I believe the current direction is correct. `remote_size` is the available number of remote block IDs, and`remote_end` is the exclusive upper bound of the remote index we may access. So the safety condition should be`remote_size >= remote_end`.

Changing it to `CHECK_GE(remote_end, remote_size)` would reject valid cases where the remote side has more blocks than this step needs.

shifengmin (dismissed):

shifengmin (approved):

XuZhang99 (approved):

PR #1541, opened 2026-05-25
[WIP]perf: overlap mtp draft extend preparation

pjgao:
## Summary

This PR overlaps the lightweight MTP draft-extend input preparation with target-model validation.

The change is intentionally conservative: it does **not** pre-dispatch draft computation and does **not** write draft KV before target validation finishes. It only moves target-output-independent preparation into the window after `impl_->step_async(validate_input)` and before `future.get()`.

## Optimization Idea

This follows the xLLM NPU optimization skill workflow:

1. Profile first and identify the MTP draft/validate bubble.
2. Separate work that depends on target validation output from deterministic host-side preparation.
3. Implement the smallest equivalent overlap first.
4. Validate both performance and accuracy before submitting.

In `MTPWorkerImpl::run_validate()`:

- Target validate is launched asynchronously.
- While target runs on device, xLLM prepares a `DraftExtendPrepareContext`.
- After target validate returns, the real accepted prefix is used to finish the draft extend input.
- The draft model is then executed exactly as before.

Prepared early:

- `base_input.to(device_, dtype_)`
- CPU views for `token_ids`, `positions`, `block_tables`
- input token ids
- sequence count, token count, block size, chunked-prefill mode

Still delayed until target validation completes:

- accepted-token scan
- actual `last_idx`
- target embeddings
- position/cache-slot/seq-len construction for the accepted prefix
- draft model dispatch
- embedding cache write

## Important Negative Result

I also tested a more aggressive P0 version that enumerated all possible `last_idx` candidates before target validation completed. That was equivalent but slower because it introduced extra CPU/vector construction work for candidates that are never used.

So this PR keeps only the low-risk P0b version.

## Benchmark

Environment:

- Model: Qwen35-27B
- Draft model: Qwen35-27B-mtp
- Hardware: Ascend 910B3 A3, Phy 8-11
- TP: 4
- MTP: `--num_speculative_tokens 3`
- Chunk prefill: enabled
- Dataset: random 20k input / 1k output
- EvalScope: `parallel=1`, `number=5`

| Version | Avg Latency(s) | TTFT(ms) | TPOT(ms) | Output TPS | Decoded/Iter | Accept |
|---|---:|---:|---:|---:|---:|---:|
| transpose-opt baseline | 14.03 | 2471.5 | 11.57 | 69.31 | 3.13 | 68.0% |
| P0 enumerate candidates | 14.81 | 2358.6 | 12.46 | 65.75 | 2.90 | 65.5% |
| P0b light prepare | 13.85 | 2466.0 | 11.39 | 70.18 | 3.09 | 67.6% |

Compared with transpose-opt baseline:

- Output TPS: `69.31 -> 70.18` (`+1.3%`)
- TPOT: `11.57 ms -> 11.39 ms` (`-1.6%`)
- Avg latency: `14.03 s -> 13.85 s` (`-1.3%`)

Artifacts:

- Benchmark: `/home/g00510989/xllm/runs/20260525_mtp3_async_draft_p0b/perf/random20k_1k_p1_n5/20260525_072650/Qwen35-27B/performance_summary.txt`
- Analysis report: `/home/g00510989/xllm/runs/20260525_mtp3_async_draft_p0b/analysis/mtp_draft_extend_prepare_overlap_report.md`

## Accuracy

GSM8K smoke test:

- `limit=10`
- `mean_acc=1.0`

Artifact:

- `/home/g00510989/xllm/runs/20260525_mtp3_async_draft_p0b/accuracy/gsm8k_limit10/reports/Qwen35-27B/gsm8k.json`

## Tests

- `sampler_test`: 11 passed, 2 skipped
- EvalScope random 20k/1k benchmark passed
- EvalScope GSM8K `limit=10` accuracy passed

## Follow-up

The larger vLLM-style async draft-dispatch optimization likely requires a shadow/pending draft KV state or double-buffered KV with commit/rollback after target validation. This PR is the safe first step and keeps the draft KV semantics unchanged.

### DIFF
--- a/xllm/core/runtime/mtp_worker_impl.cpp
+++ b/xllm/core/runtime/mtp_worker_impl.cpp
@@ -571,6 +571,13 @@ std::optional<ForwardOutput> MTPWorkerImpl::run_validate(
   Timer timer;
   timer.reset();
   auto future = impl_->step_async(validate_input);
+
+  DraftExtendPrepareContext draft_extend_context;
+  if (!input.input_params.embedding_ids.empty()) {
+    prepare_draft_extend_context(
+        input, options_.num_speculative_tokens() + 1, draft_extend_context);
+  }
+
   ForwardOutput target_output = std::move(future).get().value();
   COUNTER_ADD(speculative_execution_latency_seconds_target,
               timer.elapsed_seconds());
@@ -584,7 +591,11 @@ std::optional<ForwardOutput> MTPWorkerImpl::run_validate(
 
   // update cache and clear embeddings
   timer.reset();
-  run_draft_extend(input, val_output);
+  if (draft_extend_context.prepared) {
+    run_draft_extend(input, val_output, draft_extend_context);
+  } else {
+    run_draft_extend(input, val_output);
+  }
   COUNTER_ADD(speculative_execution_latency_seconds_draft,
               timer.elapsed_seconds());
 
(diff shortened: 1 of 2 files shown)


pjgao:
Addressed the Gemini review comments in `b07851c3`.

Changes:

- Removed the TileLang patch hunk that changed `make -j${MAKE_JOBS}` to unlimited `make -j`, so the build keeps bounded parallelism.
- Cleared `conv_weight_transposed_` when `conv1d.weight` is reloaded in `load_common_state_dict()`, avoiding stale cached weights.
- Replaced shape-based layout detection in `process_mixed_qkv()` with an explicit `input_is_btc_format` argument. The spec-verify path passes `true`; other paths keep the existing transpose.

Validation:

- `cmake --build /home/g00510989/xllm/xllm/build/cmake.linux-aarch64-cpython-311 --target sampler_test xllm -j 16`: passed
- `/home/g00510989/xllm/xllm/build/lib.linux-aarch64-cpython-311/xllm/sampler_test`: 11 passed, 2 skipped

pjgao:
Updated the PR branch to minimize the diff against the latest `preview/qwen3.5-qwen3.6` base (`c60555bf`).

Current PR head:

- `eaff9517 perf: overlap mtp draft extend preparation`

Current diff only touches:

- `xllm/core/runtime/mtp_worker_impl.cpp`
- `xllm/core/runtime/mtp_worker_impl.h`

The previous TileLang revert / TileLang patch / Qwen3 transpose / beam_search changes are no longer part of this PR branch. Those changes are either already handled by the latest preview branch or unrelated to this draft-extend prepare overlap.

Local checks:

- `git diff --check origin/preview/qwen3.5-qwen3.6..HEAD`: passed
- Verified the minimized diff is 1 commit / 2 files.

Build note:

- The earlier non-minimized worktree build passed `cmake --build ... --target sampler_test xllm -j 16` and `sampler_test` (11 passed, 2 skipped).
- A fresh isolated worktree build was attempted after minimization, but the first-time CMake setup was blocked by local dependency setup issues (`libtorch-src` FetchContent extraction directory already existed after a retry). This is an environment/bootstrap issue rather than a compile error in the minimized runtime files.

PR #1640, opened 2026-06-03
feat: support pause and resume for fully async-rl.

Clement-Wang26:
## Description

Add native pause/resume APIs to support fully asynchronous RL training, and
rename the existing weight-transfer "D2D" naming to "P2P" to better reflect its
cross-node/cross-process nature.

### 1. Pause / Resume for async RL (vLLM-compatible)

In async RL, the inference engine must periodically stop generation so the
trainer can sync updated weights, then resume. This PR adds `pause`/`resume` to
the scheduler, exposed both as engine-level methods and as HTTP endpoints
(`POST /pause`, `POST /resume`).

Three pause modes are supported, matching vLLM semantics:

| Mode | In-flight requests | KV cache | Client impact |
|------|--------------------|----------|---------------|
| `keep` (default) | preempted and pushed back to the waiting queue; re-prefilled on resume | freed | none, requests continue after resume |
| `abort` | cancelled and returned to the client | freed | client must retry |
| `wait` | allowed to finish naturally before pausing | preserved | none |

Key implementation points:
- Pause state is tracked in the scheduler (`pause_state_`, atomic) and the pause
  check sits at the very top of `ContinuousScheduler::step()`, so it is
  inherited by all schedulers that delegate to it (ChunkedPrefill, PrefillOnly,
  Mix, ZeroEviction, and the PD schedulers that call `ContinuousScheduler::step`).
- The `PAUSING -> PAUSED` transition is driven from inside the scheduler loop
  thread (`try_complete_pause`). For `enable_schedule_overlap`, the in-flight
  forward batch (`last_batch_`) is drained first so KV cache is not freed while
  a forward is still reading/writing it — this prevents garbled output after
  resume.
- `POST /pause` blocks until the scheduler has actually reached `PAUSED`
  (`wait_until_paused`, condition-variable based), so when it returns it is safe
  to update weights. This matches vLLM's behaviour where the API server only
  returns once the engine is truly paused.

### 2. Rename D2D -> P2P

The weight-transfer link APIs (`link_d2d`/`unlink_d2d`, `LinkD2D`/`UnlinkD2D`,
`D2DLinkRequest`, etc.) are renamed to the `P2P` equivalents across the proto
definitions, RPC services, scheduler, engine, worker, and HTTP routes. This is a
pure rename with no behavioural change.

> Note: this changes the HTTP routes `/link_d2d` -> `/link_p2p` and
> `/unlink_d2d` -> `/unlink_p2p`.

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Performance improvement
- [x] Refactor
- [ ] Documentation
- [x] Test
- [ ] Build or CI

## Pull Request Checklist

### Pre-commit Checks

- [ ] I have installed `pre-commit` by running `pip install pre-commit` or an equivalent command.
- [ ] I have installed the hooks with `pre-commit install`.
- [ ] I have run `pre-commit run --all-files` and fixed any reported issues.

### Self Review

### Build and Test Coverage

## Reviewer Notes

- **Scope of pause/resume:** verified to take effect for single-instance
  deployments (including TP/DP/EP multi-rank within one engine, which is driven
  by a single scheduler). PD-disaggregated deployments are **not** covered yet —
  in P/D mode a `keep` preempt on the decode side can strand requests (decode
  cannot self-prefill, and prefill no longer re-sends), similar to the DPEP
  deadlock vLLM describes. Cross-instance two-phase coordination for P/D is left
  as a follow-up.
- **Tests:** unit tests cover `keep`/`abort` preemption (KV freed, requests
  re-queued / cancelled), pause/resume state-machine transitions and
  idempotency, and re-scheduling of preempted requests. They run via
  `prepare_batch_test` and therefore do not exercise the `enable_schedule_overlap`
  drain path or the blocking `/pause` HTTP path — those were validated manually
  on NPU (Qwen3-8B): KEEP pause/resume mid-decode produces coherent output
  (earlier garbled-output bug fixed by draining the in-flight pipeline first).
- **Schedulers with a custom `step()`** (`DisaggPDScheduler`, `PDOOCScheduler`)
  still delegate to `ContinuousScheduler::step()`, so the pause check applies;
  `FixedStepsScheduler` does not delegate and is not covered.

### DIFF
--- a/tests/core/scheduler/continuous_scheduler_test.cpp
+++ b/tests/core/scheduler/continuous_scheduler_test.cpp
@@ -875,6 +875,159 @@ TEST(ContinuousSchedulerTest, PDDecodeBestOfOneSkipsExpansionAndShares) {
   // prepare_batch may allocate extra decode blocks; release them before engine
   // teardown (BlockManagerImpl checks all blocks are on the free list).
   block_manager_pool->deallocate_without_cache(seq0);
+  (void)engine.release();
+}
+// ============== Async RL training: Pause/Resume tests ==============
+
+// TEST: pause()/resume() state transitions are correct and idempotent.
+TEST(ContinuousSchedulerTest, PauseResumeStateTransition) {
+  int block_num = 9;
+  int block_size = 32;
+  ContinuousScheduler::Options opt =
+      create_scheduler_options(10000, 256, 0, 1024, 1);
+  auto engine = std::make_unique<FakeEngine>(block_num, block_size);
+  auto scheduler = std::make_unique<ContinuousScheduler>(engine.get(), opt);
+
+  EXPECT_FALSE(scheduler->is_paused());
+  scheduler->pause();
+  EXPECT_TRUE(scheduler->is_paused());
+  scheduler->pause();  // idempotent
+  EXPECT_TRUE(scheduler->is_paused());
+  scheduler->resume();
+  EXPECT_FALSE(scheduler->is_paused());
+  scheduler->resume();  // idempotent
+  EXPECT_FALSE(scheduler->is_paused());
+
+  (void)engine.release();
+}
+
+// TEST: pause preempts all running requests, frees KV cache, moves them back
+// to the waiting queue (vLLM-compatible semantics for RL).
+TEST(ContinuousSchedulerTest, PausePreemptsRunningAndFreesKVCache) {
+  int block_num = 33;
+  int block_size = 32;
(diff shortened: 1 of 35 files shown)


shifengmin (dismissed):

yq33victor (dismissed):
LGTM

shifengmin (approved):

yq33victor (approved):
LGTM

PR #1737, opened 2026-06-15
feat: replace conv1d Triton operator with AscendC operator for Qwen3.5.

maojunx99:
## Description
- Add causal_conv1d_out API with pre-allocated output for graph task update
- Introduce AclGraphTaskUpdateContext to record and replay conv1d tasks
- Update AclGraph capture/replay to handle conv1d task parameter updates
- Add is_hybrid_linear_attention() model interface for Qwen3.5

Performance :
1. Average TPOT reduced by 10%+. 
2. Qwen3.5 27B TP4 MTP: 20k input + 1k output && 1 batch, avg tpot 9.7ms.

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Performance improvement
- [ ] Refactor
- [ ] Documentation
- [ ] Test
- [ ] Build or CI

### Build and Test Coverage

### DIFF
--- a/xllm/core/framework/model/causal_vlm.h
+++ b/xllm/core/framework/model/causal_vlm.h
@@ -61,6 +61,14 @@ class CausalVLMImpl : public CausalVLM {
     return model_->forward(tokens, positions, kv_caches, parameters);
   }
 
+  bool is_hybrid_linear_attention() override {
+    if constexpr (detail::has_is_hybrid_linear_attention<Model>::value) {
+      return model_->is_hybrid_linear_attention();
+    } else {
+      return CausalLM::is_hybrid_linear_attention();
+    }
+  }
+
   torch::Tensor pooler(const torch::Tensor& hidden_states,
                        const torch::Tensor& seleted_idxes) override {
     if constexpr (detail::has_pooler<Model>::value) {
--- a/xllm/core/layers/npu_torch/qwen3_gated_delta_net_base.cpp
+++ b/xllm/core/layers/npu_torch/qwen3_gated_delta_net_base.cpp
@@ -378,49 +378,6 @@ torch::Tensor run_causal_conv1d_graph_update(
   return output;
 }
 
-torch::Tensor run_spec_verify_conv(const torch::Tensor& mixed_qkv,
-                                   const torch::Tensor& conv_cache,
-                                   const torch::Tensor& logical_state_indices,
-                                   const torch::Tensor& num_accepted_tokens,
-                                   const torch::Tensor& q_cu_seq_lens,
-                                   const torch::Tensor& conv_weight,
-                                   int32_t conv_kernel_size) {
-  const int64_t batch_size = mixed_qkv.size(0);
-  const int64_t seq_len = mixed_qkv.size(2);
-  const int64_t expanded_state_len = conv_cache.size(1);
-  CHECK_EQ(expanded_state_len, conv_kernel_size - 1 + seq_len - 1)
-      << "unexpected speculative conv cache len, expected "
-      << (conv_kernel_size - 1 + seq_len - 1) << ", got " << expanded_state_len;
-
-  xllm::kernel::CausalConv1dUpdateParams conv1d_params;
-  conv1d_params.x = mixed_qkv.transpose(1, 2)
-                        .reshape({batch_size * seq_len, mixed_qkv.size(1)})
-                        .contiguous();
-  conv1d_params.conv_state = conv_cache.transpose(1, 2);
-  conv1d_params.weight = conv_weight;
-  conv1d_params.activation = true;
-  conv1d_params.conv_state_indices =
-      expand_sequence_tensor_to_batch(
-          logical_state_indices, batch_size, "logical_state_indices")
-          .contiguous();
-  conv1d_params.num_accepted_tokens =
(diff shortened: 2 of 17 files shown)


xllm/core/runtime/acl_graph_executor_impl.cpp
```diff
@@ -69,7 +67,8 @@ bool AclGraph::capture(CausalLM* model,
                        const torch::Tensor& positions,
                        const ModelInputParams& params,
                        std::vector<KVCache>& kv_cache,
-                       uint32_t bucket_num_tokens) {
+                       uint32_t bucket_num_tokens,
+                       const ModelArgs& args) {
```
zhang-minchao: The newly added `const ModelArgs& args` parameter to `AclGraph::capture()` is never used in the function body — every branch inside `capture()` now decides behavior via `model->is_hybrid_linear_attention()` instead of the model type. This leaves `args` as a dead parameter.

Suggest removing it from the `capture()` signature (here and in `acl_graph_executor_impl.h`) and dropping the `args_` argument at the call site in `AclGraphExecutorImpl::run()`, to avoid the unused parameter and the extra coupling to `ModelArgs`.

xllm/core/layers/npu_torch/qwen3_gated_delta_net_base.cpp
```diff
@@ -316,6 +317,67 @@ torch::Tensor expand_sequence_tensor_to_batch(const torch::Tensor& tensor,
...
+  task.pad_slot_id = xllm::npu::kCausalConv1dGraphPadSlotId;
+  task.run_mode = 1;
+  task.branch = branch;
+  task.handle = handle;
+  task.event = std::move(event);
+  graph_context->causal_conv1d_tasks.emplace_back(std::move(task));
+  return output;
+}
+
 torch::Tensor run_spec_verify_conv(const torch::Tensor& mixed_qkv,
```
zhang-minchao: After this PR moves the spec-verify branch to `causal_conv1d_out` / `run_causal_conv1d_graph_update`, `run_spec_verify_conv` no longer has any caller (a full-file search finds only this definition). It was also the only remaining user of `xllm::kernel::causal_conv1d_update` in this file.

Since this PR replaces the old conv1d path, suggest removing this now-dead function to keep the diff self-contained. If it is intentionally kept as a fallback, please note that in the PR description.

xllm/core/layers/npu_torch/qwen3_gated_delta_net_base.cpp
```diff
@@ -573,37 +638,50 @@ torch::Tensor Qwen3GatedDeltaNetBaseImpl::forward(
...
-    mixed_qkv = xllm::kernel::causal_conv1d_update(conv1d_params);
-    // Reshape back to 3D [batch_size, dim, seq_len]
-    mixed_qkv =
-        mixed_qkv.view({batch_size, -1, mixed_qkv.size(-1)}).contiguous();
+    if (use_spec_verify) {
+      CHECK(input_params.num_accepted_tokens.defined())
+          << "num_accepted_tokens must be populated for Qwen3.5 spec verify";
+    }
+    torch::Tensor conv_input = reshape_qkvz_unpad(attn_metadata, mixed_qkv);
+    const auto& num_accepted = use_spec_verify
```
yingxudeng, line 604: Did you verify accuracy for the ACL graph off / MTP on case? Is it normal?

xllm/core/framework/model/causal_vlm.h
```diff
@@ -61,6 +61,14 @@ class CausalVLMImpl : public CausalVLM {
     return model_->forward(tokens, positions, kv_caches, parameters);
   }
 
+  bool is_hybrid_linear_attention() override {
```
yingxudeng, line 64: Has image inference been tested and verified to work correctly?

xllm/core/layers/npu_torch/qwen3_gated_delta_net_base.cpp
yingxudeng: It switches from ACL graph to eager mode when batch > 16. This might be problematic for online inference.

maojunx99: Even without this PR, the accuracy for the ACL graph off / MTP on case is already abnormal. It needs further fixes in follow-up work.

xllm/core/framework/model/causal_vlm.h
maojunx99: Multimodal accuracy has been verified.

xllm/core/layers/npu_torch/qwen3_gated_delta_net_base.cpp
maojunx99: Done. "run_spec_verify_conv" has been removed.

xllm/core/runtime/acl_graph_executor_impl.cpp
maojunx99: Done. Removed the unused ModelArgs& args parameter from AclGraph::capture() signature and dropped the args_ argument at the call site in AclGraphExecutorImpl::run().

maojunx99:
local test results:
image

yingxudeng (approved):

PR #1743, opened 2026-06-15
feat: support in-batch prefix cache.

Clement-Wang26:
## Description

Support **in-batch prefix cache**: publish the full prefill blocks admitted
within a single scheduling step into the prefix cache immediately, so that later
requests in the *same* batch sharing the same prefix can reuse them — instead of
waiting until a request finishes its own prefill (first decode step) or
deallocates. This closes the gap where several concurrent same-prefix requests
in one prefill batch each recompute the shared prefix.

Key points:
- New flag `enable_in_batch_prefix_cache` (default `false`), wired through
  `KVCacheConfig` and the scheduler options. It is an opt-in enhancement layered
  on top of `enable_prefix_cache`; when prefix cache is off, this feature is a
  no-op by design (guarded at scheduler, `BlockManagerPool`, and
  `BlockManagerImpl` levels).
- Add `KVCacheManager::cache(Sequence*, size_t num_tokens)` overload that
  publishes only the full blocks covered by the step's token budget (clamped to
  allocated blocks and the sequence's own tokens; partial trailing block is
  dropped by the prefix-cache insert). This differs from the existing
  `cache(Sequence*)`, which caches already-computed tokens at deallocate time.
- Bound the hierarchy offload range by the computed device blocks so that
  blocks registered early (but not yet computed) are not offloaded to host.
- Invoke `cache_in_batch_prefix` per admitted request in the continuous,
  chunked-prefill, mix, and prefill-only schedulers. In chunked prefill it runs
  per chunk and incrementally publishes only the newly completed full blocks.

## Related Issues

N/A

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Performance improvement
- [ ] Refactor
- [ ] Documentation
- [ ] Test
- [ ] Build or CI

### Pre-commit Checks

### Self Review

### Build and Test Coverage

## Reviewer Notes

- Correctness relies on the invariant that **prefill sequences admitted in a
  step are not preempted within that same step** (so the blocks published before
  forward are guaranteed to be computed by that step's forward). This holds today
  because decode-side preemption only runs when no prefill is scheduled
  (`if (running_sequences_.empty())`), and prefill-side preemption only evicts
  previously-running offline requests. If the scheduler changes here, revisit
  this feature.
- Default is `false`; behavior is unchanged unless the flag is explicitly enabled
  on top of `enable_prefix_cache`.
- Not applied to `ZeroEvictionScheduler` / `FixedStepsScheduler` (they override
  `handle_prefill_requests` without the call) nor to
  `DisaggPDChunkedPrefillScheduler`'s custom chunked path.
- Validation gap: a full `python setup.py build test` could not be completed in
  the current environment (unrelated mid-refactor proto staleness + an
  `xllm_ops` reconfigure failure). The new scheduler test TU was verified to
  compile cleanly against current headers; please run the test suite on a proper
  build machine before merge.

### DIFF
--- a/tests/core/framework/block/block_manager_test.cpp
+++ b/tests/core/framework/block/block_manager_test.cpp
@@ -446,4 +446,29 @@ TEST(BlockManagerPoolTest, SequenceUpdateBlockHashes) {
             0);
 }
 
+// In-batch prefix cache publishes only the full blocks covered by the given
+// token budget. When the budget is overestimated, cache() must clamp to the
+// sequence's own tokens and register just the complete blocks.
+TEST(BlockManagerPoolTest,
+     CachePrefixClampsToSequenceTokensWhenBudgetIsOverestimated) {
+  ScopedValue<int32_t> max_seqs_guard(
+      &SchedulerConfig::get_instance().max_seqs_per_batch(), 2);
+
+  BlockManagerPool::Options options;
+  options.num_blocks(16).host_num_blocks(0).block_size(4).enable_prefix_cache(
+      true);
+  // Leak the pool intentionally: with prefix cache enabled, the cached block
+  // stays referenced by the prefix-cache table at teardown, which would trip
+  // the free-list check in ~BlockManagerImpl.
+  auto* pool = new BlockManagerPool(options, /*dp_size=*/1);
+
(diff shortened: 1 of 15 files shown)


shifengmin (approved):

yq33victor (approved):
LGTM

Kang-Meng (approved):

PR #1776, opened 2026-06-18
bugfix: resolve cross-NUMA spawn worker isolation issues

asr-sheep1:
## Description
This PR fixes several issues in the existing cross-NUMA spawn worker implementation on CUDA/MLU/DCU, including:

- Fix PCI bus_id case mismatch for NUMA detection (observed on CUDA).

- Initialize spawn_worker_dir for the xllm server binary and add the --spawn_worker_dir option.

- Reset inherited CPU affinity and NUMA memory policy before binding the spawned worker to the target NUMA node.
  (This change saves the original CPU affinity before engine NUMA binding, clears the memory policy inherited by spawned workers, restores the saved affinity in each spawned worker, and then binds the worker to its target NUMA node.)

These fixes improve the correctness of spawn worker initialization and NUMA binding. While no fatal issues were observed in small-scale model testing, they eliminate several potential issues in cross-NUMA deployments.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Refactor
- [ ] Documentation
- [ ] Test
- [ ] Build or CI

## Pull Request Checklist

### Pre-commit Checks

### Self Review

### Build and Test Coverage

- [x] CUDA: `python setup.py build test` has passed on a CUDA machine.

### DIFF
--- a/xllm/core/platform/numa_utils.cpp
+++ b/xllm/core/platform/numa_utils.cpp
@@ -59,10 +59,11 @@ int32_t get_numa_node_from_sysfs(const std::string& backend,
                                  const std::string& pci_bus_id) {
   // Device APIs may return uppercase hex, but sysfs PCI dirs require lowercase.
   std::string lower_bus_id = pci_bus_id;
-  std::transform(lower_bus_id.begin(),
-                 lower_bus_id.end(),
-                 lower_bus_id.begin(),
-                 [](unsigned char ch) { return std::tolower(ch); });
+  std::transform(
+      lower_bus_id.begin(),
+      lower_bus_id.end(),
+      lower_bus_id.begin(),
+      [](unsigned char ch) { return static_cast<char>(std::tolower(ch)); });
   std::string numa_path = "/sys/bus/pci/devices/" + lower_bus_id + "/numa_node";
   LOG(INFO) << backend << " device " << device_index
             << " NUMA sysfs path: " << numa_path;
--- a/xllm/core/platform/numa_utils.cpp
+++ b/xllm/core/platform/numa_utils.cpp
@@ -28,6 +28,8 @@ limitations under the License.
 #include <sched.h>
 #include <unistd.h>
 
+#include <algorithm>
+#include <cctype>
 #include <cerrno>
 #include <cstring>
 #include <fstream>
@@ -55,7 +57,13 @@ bool read_numa_node(const std::string& numa_path, int32_t* numa_node) {
 int32_t get_numa_node_from_sysfs(const std::string& backend,
                                  int32_t device_index,
                                  const std::string& pci_bus_id) {
-  std::string numa_path = "/sys/bus/pci/devices/" + pci_bus_id + "/numa_node";
+  // Device APIs may return uppercase hex, but sysfs PCI dirs require lowercase.
+  std::string lower_bus_id = pci_bus_id;
+  std::transform(lower_bus_id.begin(),
+                 lower_bus_id.end(),
+                 lower_bus_id.begin(),
+                 [](unsigned char ch) { return std::tolower(ch); });
+  std::string numa_path = "/sys/bus/pci/devices/" + lower_bus_id + "/numa_node";
   LOG(INFO) << backend << " device " << device_index
             << " NUMA sysfs path: " << numa_path;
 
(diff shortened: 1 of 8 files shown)


xllm/xllm.cpp
```diff
@@ -234,6 +234,8 @@ Options create_options(const std::string& instance_name, bool is_local) {
     options.draft_devices(draft_devices);
   }
 
+  options.spawn_worker_path(execution_config.spawn_worker_dir());
```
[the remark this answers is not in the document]
asr-sheep1: `options.spawn_worker_path()` is treated as a **directory** in the subsequent code (see `worker_server.cpp:285`), where the `/spawn_worker` suffix is appended automatically.

Although the name is somewhat misleading, its intended semantics are "a directory" rather than "the full executable path". Appending the suffix when assigning the value would cause it to be appended twice, resulting in a startup failure.

Renaming it would require additional changes to the existing variable naming and related code. To minimize the scope of this PR, the existing naming is preserved.

xllm/core/platform/numa_utils.cpp
```diff
@@ -222,6 +230,34 @@ int32_t get_device_numa_node(int32_t device_index) {
...
+  }
+
+  // Reset to all CPUs so build_cpu_set_for_numa_node can pick the target
+  // NUMA's CPUs (inherited mask would leave the intersection empty).
+  cpu_set_t all_cpus;
+  CPU_ZERO(&all_cpus);
+  const int32_t nr_possible_cpus = numa_num_possible_cpus();
+  for (int32_t cpu = 0; cpu < nr_possible_cpus && cpu < CPU_SETSIZE; ++cpu) {
+    CPU_SET(cpu, &all_cpus);
+  }
```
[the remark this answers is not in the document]
asr-sheep1: The return value is guaranteed to be positive by the is_numa_available() check at the beginning of the function.

xllm/xllm.cpp
weizhehuang0827, line 233: The `spawn worker` should only be used in offline inference scenarios (Python interface) and should not appear in `xllm.cpp` (which is the entry point for online inference). I don't know why this `gflags` is added.

asr-sheep1: I think I understand your point now.

In `dist_manager.cpp`, `use_spawn_worker` is currently controlled by:
https://github.com/jd-opensource/xllm/blob/7f33bde1362bb00735666607d686aba7856aa20b/xllm/core/distributed_runtime/dist_manager.cpp#L250 `bool use_spawn_worker = (options.enable_offline_inference() && i > 0) || force_spawn_for_numa_isolation[i];`

So in online serving, if the local devices span multiple NUMA nodes, `force_spawn_for_numa_isolation[i]` may still make the worker enter the `spawn_worker` path. Since `xllm.cpp` does not set `spawn_worker_path`, this can make `spawn_worker` launch fail and cause startup to hang.

So adding `spawn_worker_dir` to `xllm.cpp` was not the right fix. The root issue seems to be the boundary condition for `spawn_worker` usage: `spawn_worker` should stay scoped to offline inference and should not be forced by the NUMA isolation logic in online serving.

Therefore, I’m inclined to first remove the `spawn_worker_dir` changes, and then focus the fix on the `spawn_worker` boundary condition. Does this understanding and direction look right to you?

asr-sheep1: Sorry, my previous reply was left pending and wasn’t submitted in time. @weizhehuang0827 , could you please take another look? I’ll update the commits and rebase the branch if any changes are needed.

weizhehuang0827: Do not add a `spawn_worker_dir` flag, and the `spawn worker` function is only for offline Python inference. If you don't need this, this PR is unnecessary.

weizhehuang0827: If your NUMA core binding requires this `spawn_worker_dir`, you should consider adding or handling it in the Python interface.

asr-sheep1: I understand your point now.  However, our CUDA tests show that online inference with **multiple GPUs across multiple NUMA nodes still takes the spawn-worker path**. due to the logic here:
https://github.com/xLLM-AI/xllm/blob/cd301c336d44e7d54a986c2c5883efae69b2ff70/xllm/core/distributed_runtime/dist_manager.cpp#L252

I think use_spawn_worker should also be gated by enable_offline_inference():
`bool use_spawn_worker =
    options.enable_offline_inference() &&
    (i > 0 );`

weizhehuang0827: The `use_spawn_worker` function should already be subject to the restrictions of `enable_offline_inference()`, so it seems there might be a problem with the logic that was modified earlier.

asr-sheep1: I understand now. I will rework this PR so spawn_worker remains strictly scoped to offline Python inference: remove the spawn_worker_dir flag and xllm.cpp wiring, and fix the DistManager boundary condition.

PR #1783, opened 2026-06-18
feat: add RL deep-sleep  for co-located training.

Clement-Wang26:
## Description

Introduce a VMM-backed `SleepableAllocator` that releases and re-acquires the
physical HBM of model weights and KV cache while keeping their virtual addresses
stable, enabling vllm-ascend-style sleep mode for RLHF co-located
training/inference (`enable_sleep_mode`).

**How it works**
- **SleepableAllocator**: owns per-tag VMM regions (`weights` / `kv_cache`).
  `sleep()` unmaps the physical memory (`aclrtFreePhysical`) and returns it to
  the driver; `wake_up()` re-maps fresh physical memory to the *same* virtual
  address, so tensors built over it stay valid (contents are discarded — deep
  sleep only, matching vllm-ascend level 2).
- **KV cache**: in sleep mode the per-layer K/V tensors are built directly over a
  VMM region via `get_tensor_from_blob`, so the whole KV cache is released/
  re-acquired in one shot.
- **Weights**: the manual loader allocates each decoder layer's contiguous
  weight buffer from the `WEIGHTS` region, so weights become sleepable without
  any allocator routing (torch_npu's pluggable allocator is only reachable from
  Python, not from xLLM's C++ load path).
- **update_weights**: reload weights in place from disk (or a new checkpoint)
  after `wake_up`, for RLHF weight sync.
- Wire `sleep` / `wake_up` / `is_sleeping` / `update_weights` through engine,
  master, worker RPC and the Python `LLM` offline API. The RL path is kept fully
  isolated from the existing xtensor (`PageAllocator`) sleep/wakeup path.

**Why not the torch_npu pluggable allocator (like vllm-ascend)?**
vllm-ascend wraps weight loading / KV creation in `use_memory_pool(allocator)`,
which only works because both happen in Python on the same thread. In xLLM those
allocations happen in C++ worker threads, and `MemPool(custom_allocator)` /
`createCustomAllocator` are not exported as C++ symbols (only via pybind), so the
pool cannot be backed by a custom allocator from C++. Building tensors directly
over VMM regions achieves the same outcome (only weights + KV are sleepable;
everything else stays on the default allocator).

## Related Issues

N/A

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Performance improvement
- [x] Refactor
- [ ] Documentation
- [ ] Test
- [ ] Build or CI

### Pre-commit Checks

### Self Review

### Build and Test Coverage


xllm/core/framework/kv_cache/kv_cache.cpp
```diff
@@ -89,6 +90,104 @@ std::string int32_vector_string(const std::vector<int32_t>& values) {
...
+  for (int64_t d : shape) {
+    n *= static_cast<size_t>(d);
+  }
+  return n;
+}
+
+// Wrap a raw device address (inside a SleepableAllocator region) as a
+// torch::Tensor with the given shape/dtype, without owning the memory. Mirrors
+// XTensor's NPU storage construction; falls back to from_blob elsewhere.
+torch::Tensor make_region_tensor(void* addr,
```
XuZhang99: move `numel_of` and `make_region_tensor` to `tensor_helper.h`

xllm/core/platform/vmm_api.cpp
```diff
@@ -140,6 +140,58 @@ void create_phy_mem_handle(PhyMemHandle& phy_mem_handle, int32_t device_id) {
       granularity_size);
 }
 
+void create_phy_mem_handle(PhyMemHandle& phy_mem_handle,
+                           int32_t device_id,
+                           size_t size) {
```
XuZhang99, line 101: what is the difference between this func and `void create_phy_mem_handle(PhyMemHandle& phy_mem_handle, int32_t device_id);`，`granularity_size` is fixed to `2MB` and has no other value.

xllm/core/platform/vmm_api.cpp
```diff
@@ -272,5 +324,27 @@ void unmap(VirPtr& vir_ptr, size_t aligned_size) {
 #endif
 }
 
+void unmap_chunk(VirPtr& vir_ptr, size_t size) {
```
XuZhang99, line 304: what's the difference from `void unmap(VirPtr& vir_ptr, size_t aligned_size);`?

xllm/core/framework/kv_cache/kv_cache.cpp
Clement-Wang26: Good catch — make_region_tensor duplicated get_tensor_from_blob in util/tensor_helper.h, so I removed it and reuse that helper. numel_of is now a one-line std::accumulate, so I didn't add a separate helper.

xllm/core/platform/vmm_api.cpp
Clement-Wang26: The 2-arg create_phy_mem_handle(handle, device_id) always allocates a single granularity-sized (2MB) page and writes the global KVCacheConfig::phy_page_granularity_size (used by the xtensor per-page map/unmap). The size-aware overload allocates an arbitrary-size handle with no global side effect, which lets SleepableRegion map the KV/weight regions as a few large chunks (1GiB / per-layer) instead of thousands of 2MB pages. I've refactored the 2-arg version to delegate to the size version, so they no longer duplicate the per-platform allocation code.

xllm/core/platform/vmm_api.cpp
Clement-Wang26: unmap(ptr, aligned_size) is the xtensor path: it loops over the range in phy_page_granularity_size (2MB) steps calling aclrtUnmapMem once per page, because that path maps memory page-by-page. unmap_chunk(ptr, size) issues a single aclrtUnmapMem(ptr) for a chunk that was mapped by one aclrtMapMem call covering the whole chunk (how SleepableRegion maps). On NPU this distinction matters because aclrtUnmapMem unmaps exactly what one aclrtMapMem mapped; on CUDA/MLU/DCU both are equivalent (cuMemUnmap(ptr, size)).

XuZhang99: for non-npu device, we can call `unmap` inside `unmap_chunk` directly.

shifengmin (dismissed):

xllm/core/platform/vmm_api.cpp
XuZhang99: fou npu, if you want to allocate `1GB`, `prop.memAttr = ACL_HBM_MEM_HUGE;  // 2MB` should be modified: https://www.hiascend.com/document/detail/zh/CANNCommunityEdition/900/API/runtimeapi/aclcppdevg_03_1398.html

for other device, also need to check the API.

also need to add `CHECK` for `size`.

we cannot simply reuse 2MB phy handle allocation func for 1GB allocation.

xllm/core/platform/vmm_api.cpp
Clement-Wang26: Done — non-NPU now delegates to unmap() (single cuMemUnmap/cnMemUnmap/… over the whole range); only NPU keeps the single aclrtUnmapMem. Also added the missing USE_MUSA branch to unmap() so the delegation is correct there.

xllm/core/platform/vmm_api.cpp
Clement-Wang26: Added CHECK(size > 0) and clarified the comment: size is the total handle size (granularity-aligned by the caller), and the 2MB huge-page attr backs arbitrary-size chunks. ACL_HBM_MEM_HUGE1G (1GiB pages, fewer page tables / better TLB) is only available on A2/A3, so I kept the portable 2MB option — consistent with the existing xtensor KV path. A HUGE1G fast-path gated on hardware capability can be a follow-up perf PR.

shifengmin (dismissed):

weizhehuang0827 (dismissed):

shanshanpt (dismissed):
LGTM

shifengmin (approved):

XuZhang99 (approved):

PR #1810, opened 2026-06-24
bugfix: restore GLM5/GLM5.2 long-context accuracy with indexer RoPE layout preprocessing

chenchuw886:
## Summary

This PR fixes GLM-5.2 NPU accuracy issues caused by the indexer RoPE layout mismatch.

For `GlmMoeDsaForCausalLM` models with `"indexer_rope_interleave": true`, the indexer RoPE inputs are expected to use the interleaved layout instead of the NeoX / half-split layout.

For example, given the input:

```text
[x0, x1, x2, x3, x4, x5, x6, x7]
```

The RoPE dimension pairing should be:

```text
NeoX / half-split:
(x0, x4), (x1, x5), (x2, x6), (x3, x7)

Interleaved:
(x0, x1), (x2, x3), (x4, x5), (x6, x7)
```

Using the wrong indexer RoPE layout changes the indexer attention scores and may lead to semantic accuracy degradation. The impact is more visible in long-context generation, where sparse index selection becomes more sensitive to positional encoding errors.

Instead of inserting extra layout conversions before or after RoPE at runtime, this PR reorders the affected indexer weights during model loading. This keeps the runtime execution path unchanged and avoids additional runtime overhead.

## Changes

* Add indexer weight preprocessing for GLM-5 / GLM-5.2 NPU models with `"indexer_rope_interleave": true`
* Reorder the RoPE-related front dimensions of:

  * `self_attn.indexer.wq_b.weight`
  * `self_attn.indexer.wk.weight`
  * associated quantization metadata for `wq_b`
* Fix `rope_theta` parsing from `config.json`
* Fix GLM-5 / GLM-5-MTP NPU model argument parsing
* Restore `numOfSharedExperts` to use the model config value instead of a hardcoded value

## Validation

* IFBench: `72.33`
* GPQA_DIAMOND: `89.39`

### DIFF
--- a/xllm/core/layers/npu/loader/deepseek_v32_decoder_loader.cpp
+++ b/xllm/core/layers/npu/loader/deepseek_v32_decoder_loader.cpp
@@ -1043,10 +1043,10 @@ void DeekseekV32DecoderLoader::preprocess_linear_for_rope() {
 
   std::vector<std::string> indexer_linear_for_rope;
   indexer_linear_for_rope.reserve(5);
-  int32_t processed_indexer_rope_tensors = 0;
   indexer_linear_for_rope.emplace_back("self_attn.indexer.wq_b.weight");
   if (is_attn_dynamic_desc(kIndexerWqBLinearIndex)) {
-    indexer_linear_for_rope.emplace_back("self_attn.indexer.wq_b.weight_offset");
+    indexer_linear_for_rope.emplace_back(
+        "self_attn.indexer.wq_b.weight_offset");
     indexer_linear_for_rope.emplace_back("self_attn.indexer.wq_b.weight_scale");
   } else {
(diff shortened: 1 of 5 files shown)


sanlio36:
1. glm5_moe.h and glm5_moe_mtp.h add nearly duplicated LOG(INFO) argument-dump helpers. These logs may be noisy in multi-rank NPU deployments and mostly duplicate existing ModelArgs visibility. Consider removing them.
2. processed_indexer_rope_tensors in deepseek_v32_decoder_loader.cpp is incremented but never used. It should be removed.
3. The extra ([&] { ... }) wrapping in REGISTER_MODEL_ARGS is unusual compared with existing style and should be simplified if not required. Please also run clang-format.

PR #1668, opened 2026-06-09
perf: overlap acl graph decode prepare with replay via double buffer.

wanghuanjun2113:
### 概要

本 PR 优化 NPU ACL Graph decode 路径，通过把下一轮 graph 输入准备和当前 graph replay 重叠起来，减少 replay 之后的 NPU 空泡。最新更新增加了独立的 ACL graph double buffer 开关，显存敏感场景可以只关闭 double buffer，退化回 single-buffer graph 路径，而不必关闭整个 graph mode 或 schedule overlap。

### Double Buffer 设计

优化前，graph replay 和下一轮输入准备容易串行化：NPU 完成 graph compute 后，还要等待 persistent input 更新，timeline 上会出现明显的 Prepare 空泡。

Double buffer 使用两个 graph slot：

- 每个 slot 独立持有一份 `GraphPersistentParam`；
- 每个 slot 独立持有按 bucket/graph key 缓存的 `AclGraph` 实例；
- replay 消费已经 prepared 的 slot，同时 prepare 更新另一个 slot；
- `GraphSlot::is_prepared` 防止覆盖还没有被 replay 消费的 prepared slot；
- `AclGraph::replay_inputs_prepared_` 继续作为 graph 内部原子交接，避免 replay 读到半准备状态。

本次更新还把普通 graph decode 的 prepare 从 compute stream 等待中解耦出来。`prepare_stream.wait_stream(compute_stream)` 只保留在 speculative decode 路径，因为 MTP 相关更新仍需要这个顺序约束。

### 显存占用分析

Double buffer 是用 graph 相关显存换 overlap。相比 single buffer，它会让 graph executor 自己持有的以下资源近似翻倍：

- `GraphPersistentParam` 中的 persistent graph 输入/输出参数；
- 每个已 capture bucket/graph key 对应的 ACL graph 实例及其 graph 侧资源。

它不会复制模型权重或 KV cache。实际额外显存取决于已 capture 的 graph bucket 数量和活跃 shape。对于大模型或 HBM 紧张场景，关闭 double buffer 可以保留 graph mode，同时把 graph persistent/captured 相关显存降回 single-buffer 级别。

### 新增开关

新增开关：

```bash
--enable_graph_double_buffer=true
```

默认值：`true`。

行为：

- `true`：使用两个 graph slot；当 `--enable_graph` 和 schedule overlap 同时开启时启用 prepared-input overlap；
[...]

PR #1854, opened 2026-06-30
fix: keep graph decode prepare independent

pjgao:
## Summary

This PR fixes a performance regression introduced by the merged PR #1668 path.

The final PR #1668 source branch was merged with a guarded wait condition in
`WorkerImpl::prepare_work_before_execute_on_stream()`:

```cpp
const bool can_prepare_graph_decode_independently =
    can_prepare_npu_graph_decode_input(input.input_params);
if (enable_schedule_overlap() &&
    (options_.enable_speculative_decode() ||
     !can_prepare_graph_decode_independently) &&
    compute_stream_) {
  prepare_stream.wait_stream(*compute_stream_);
}
```

In this prepare stage, `input.input_params` is not a reliable predicate for the
next ordinary graph-decode double-buffer prepare. As a result,
`can_prepare_graph_decode_independently` can evaluate false and ordinary
graph-decode prepare waits on the previous compute stream again. This brings
back the inter-decode host bubble that PR #1668 was intended to remove.

This PR restores the validated behavior: keep the prepare/compute stream wait
only for speculative decode, where target/draft workers share the streams and
still need the ordering. Ordinary graph-decode overlap keeps prepare independent
from compute.

## Root Cause

The reviewed and validated local fix contained:

- `0cd634f6 fix: keep graph decode prepare independent`
- `e04ac3fb fix: keep fallback token update synchronized`

However, the PR #1668 source branch that was ultimately merged pointed to
`8eb0bc52`, whose wait condition still used the
`can_prepare_npu_graph_decode_input(input.input_params)` guard. The merged
commit `5f9ec8f1` therefore reintroduced the wait for the Qwen3-1.7B graph
double-buffer decode path.

## Validation

Environment:

- Model: `Qwen3-1.7B`
- Command: `/home/g00510989/profiling/qwen1.7b/1.7/run.sh`
- Request script: `/home/g00510989/profiling/qwen1.7b/1.7/send_request.py`
- Flags: `enable_schedule_overlap=true`, `enable_graph=true`,
  `enable_graph_double_buffer=true`

Build:

- `python setup.py build` passed

Performance after this fix, `send_request.py --no-top`, 3 repeated runs:

- run1: avg TPOT `0.0049s`
- run2: avg TPOT `0.0049s`
- run3: avg TPOT `0.0049s`

Profiling after this fix:

- Boundary: `ArgMaxV2AiCore -> MODEL_EXECUTE`
- pair_count: `128`
- gap median: `8.25us`
- gap p90: `8.75us`
- gap max: `9.0us`

For comparison, current `main` before this fix showed:

- avg TPOT about `0.0054s` to `0.0055s`
- `ArgMaxV2AiCore -> MODEL_EXECUTE` gap median `706.625us`
- gap p90 `765.5us`

## Scope

This PR is intentionally limited to one runtime synchronization condition in
`worker_impl.cpp`. It does not include the local weight-loading workaround used
for this machine's validation environment.

### DIFF
--- a/xllm/core/runtime/worker_impl.cpp
+++ b/xllm/core/runtime/worker_impl.cpp
@@ -727,14 +727,10 @@ void WorkerImpl::prepare_work_before_execute_on_stream(
   }
 #endif
   c10::StreamGuard stream_guard = prepare_stream.set_stream_guard();
-  const bool can_prepare_graph_decode_independently =
-      can_prepare_npu_graph_decode_input(input.input_params);
-  if (enable_schedule_overlap() &&
-      (options_.enable_speculative_decode() ||
-       !can_prepare_graph_decode_independently) &&
+  if (enable_schedule_overlap() && options_.enable_speculative_decode() &&
       compute_stream_) {
     // MTP updates reuse shared prepare/compute streams and need this ordering;
-    // only graph double-buffer decode can prepare the next slot independently.
+    // ordinary graph-decode overlap must keep prepare independent of compute.
     prepare_stream.wait_stream(*compute_stream_);
   }
   CHECK(prepare_stream.wait_event(input.metadata_ready_event))


pjgao:
Closing this PR because the head branch was pushed from the wrong fork. Reopening from pjgao/xllm.

PR #1861, opened 2026-07-01
refactor: host prefix cache — device↔host KV-cache transfer framework (4/n).

Kang-Meng:
### Background

Previously `HierarchyBlockManagerPool` was selected when `host_blocks_factor > 1` but the
actual D2H/H2D transfer path was disabled: blocks were never offloaded to host and host
prefix-cache hits were never reloaded to device. This PR re-enables the full offload→reload
pipeline for `BlockType::KV`.

---

#### Commit 1 — allocate host KV cache as page-aligned pinned memory in KVCache impls (`ae8ba1a1`)

Ensures host-side tensors used for D2H/H2D DMA are allocated as pinned, page-aligned memory:

- `HostPageAlignedRegion`: on NPU builds, calls
  `aclrtHostRegister(..., ACL_HOST_REGISTER_MAPPED, ...)` immediately after `mmap`+`mlock`
  so the host buffer is mapped into device address space. Destructor calls
  `aclrtHostUnregister` before `munlock`/`munmap`. Move-assignment operator cleaned up (was
  leaking if `this` held existing memory).
- New utility functions in `kv_cache_utils`: `scale_host_block_count` (clamps factor to
  ≥ 1.0), `build_host_tensor_shape` (scales dim-0 by factor),
  `build_host_group_tensor_shape` (inserts a layer dimension at index 1, yielding
  `[host_blocks, layer_count, ...]`), and `create_host_page_aligned_tensor` (allocates
  region + creates `torch::from_blob` view).
- `KVCacheImpl` and all impl subclasses (Standard, DeepSeekV4, Indexed, LinearAttention)
  gain a host constructor `(KVCacheShape, KVCacheCreateOptions, BlockType, layer_count)`
  that allocates the grouped host tensor for its block type, and a
  `get_block_type_tensors(BlockType)` virtual method queried by the transfer layer.
- Unit test in `kv_cache_test.cpp` covers host tensor allocation and shape contract for
  each impl.

#### Commit 2 — restore D2H offload collection (`cbcbc2f1`)

Restores the D2H offload trigger (removed in commit 1 to keep the diff reviewable) and
fixes a race in the H2D path:

- `collect_offload_pairs()`: extracted from `deallocate()`. Allocates host blocks to match
  not-yet-offloaded device blocks (aborts without partial copy if the host pool is
  exhausted), then enqueues `OffloadBlockPair`s for blocks with `ref_count == 2` (uniquely
  owned; beam-shared blocks with `ref_count > 2` are skipped). Bounded by
  `kv_cache_tokens_num` so uncomputed in-batch blocks are never copied.
- **H2D race fix**: `LayerSynchronizer` is now created and inserted into
  `layer_wise_load_synchronizer_` **synchronously** inside `transfer_kv_blocks()`, before
  the async copy task is scheduled. Previously the synchronizer was created inside the
  threadpool task; the forward's `set_layer_synchronizer()` could look up the map before
  the task ran, miss the entry, skip the wait, and read KV before H2D completed.
- `offload_threadpool_` removed: D2H now runs synchronously on the caller (RemoteWorker
  copy thread) so the copied-block count can be returned to the scheduler. The shared
  stream pool is sized to `load_threadpool_.size() + kOffloadStreamCount (4)` to cover
  concurrent D2H callers.
- Two bounds guards on H2D copy loops: copy range clamped to
  `min(hbm_blocks.size(), host_blocks.size())` to prevent out-of-bounds when a
  chunked-prefill allocation produces fewer device blocks than the host prefix match covers.

---

### Test

Validated on A3 (Ascend 910) with Qwen3-32B TP2, `MAX_MEM_UTIL=0.70` (device pool: 480
blocks), `generated-shared-prefix` workload (60 groups × 8 prompts, 2560-token shared
prefix, concurrency=4):

|  | factor=4 (host cache) | factor=0 (device only) |
|---|---|---|
| hit_rate | 50.6% | 50.7% |
| HIT TTFT | 141 ms | 147 ms |
| MISS TTFT | **840 ms** | **974 ms** |
| Mean TTFT | 503 ms | 572 ms |
| mem-fails | 0 | 0 |

- **Activation** confirmed: `HierarchyKVCacheTransfer: created host cache for 1 block type groups.`
- **Host-backed hit** confirmed: repeated long prompt logs `num_prefix_cache_tokens: 256`,
  TTFT 74 ms → 45 ms, byte-consistent output.
- **No device-block leak**: 0 `No enough memory to schedule` events; idle recovery HTTP 200
  after full workload.
- **Accuracy**: `finish_reason=stop`, coherent French output, idle and under load.

### DIFF
--- a/xllm/core/framework/block/hierarchy_block_manager_pool.cpp
+++ b/xllm/core/framework/block/hierarchy_block_manager_pool.cpp
@@ -18,6 +18,7 @@ limitations under the License.
 #include <algorithm>
 
 #include "block_manager_impl.h"
+#include "composite_block_manager.h"
 #include "concurrent_block_manager_impl.h"
 
 namespace xllm {
@@ -31,21 +32,23 @@ HierarchyBlockManagerPool::HierarchyBlockManagerPool(
   host_block_managers_.reserve(dp_size);
 
   BlockManager::Options host_options;
-  host_options.num_blocks(options_.num_blocks())
+  host_options.num_blocks(options_.host_num_blocks())
       .block_size(options_.block_size())
       .enable_prefix_cache(options_.enable_prefix_cache())
       .enable_disagg_pd(options_.enable_disagg_pd())
-      .num_blocks(options_.host_num_blocks())
       .hasher_type(options_.hasher_type());
 
   for (int32_t i = 0; i < dp_size; ++i) {
-    if (options.enable_disagg_pd() || options_.enable_kvcache_store()) {
-      host_block_managers_.emplace_back(
-          std::make_unique<ConcurrentBlockManagerImpl>(host_options));
-    } else {
-      host_block_managers_.emplace_back(
-          std::make_unique<BlockManagerImpl>(host_options));
+    std::unique_ptr<BlockManager> leaf =
+        std::make_unique<BlockManagerImpl>(host_options);
+    // The D2H offload callback (transfer_blocks) frees host blocks from a folly
+    // executor thread while the scheduler thread also allocates/deallocates on
+    // this leaf, so it must be lock-guarded. Host offload is always on here.
+    if (options.enable_disagg_pd() || options_.enable_kvcache_store() ||
+        options_.enable_host_offload()) {
+      leaf = std::make_unique<ConcurrentBlockManagerImpl>(std::move(leaf));
     }
+    host_block_managers_.emplace_back(std::move(leaf));
   }
 
   load_block_transfer_infos_.resize(host_block_managers_.size());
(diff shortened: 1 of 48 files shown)


zhang-minchao (commented):
Review: 两个需要关注的问题

xllm/core/framework/block/hierarchy_block_manager_pool.cpp
```diff
@@ -31,21 +32,19 @@ HierarchyBlockManagerPool::HierarchyBlockManagerPool(
...
       .block_size(options_.block_size())
       .enable_prefix_cache(options_.enable_prefix_cache())
       .enable_disagg_pd(options_.enable_disagg_pd())
-      .num_blocks(options_.host_num_blocks())
       .hasher_type(options_.hasher_type());
 
   for (int32_t i = 0; i < dp_size; ++i) {
+    std::unique_ptr<BlockManager> leaf =
+        std::make_unique<BlockManagerImpl>(host_options);
     if (options.enable_disagg_pd() || options_.enable_kvcache_store()) {
```
zhang-minchao: 这里的条件只检查了 `enable_disagg_pd()` 和 `enable_kvcache_store()`，但 `enable_host_offload` 场景下 host block manager 同样存在多线程并发问题：

下面 `transfer_blocks()` 里的 offload 回调跑在 `folly::getGlobalCPUExecutor()` 线程上，会调 `host_block_mgr_ptr->cache()` 和 `->deallocate()`；与此同时 scheduler 线程会在 `collect_offload_pairs` 里调用 `host_block_managers_[dp_rank]->allocate()`。

裸的 `BlockManagerImpl` 没有内部加锁，两个线程同时操作 free_list 和 prefix-cache trie 会导致数据竞争（corruption / double-free / crash）。

建议把条件改为：
```cpp
if (options.enable_disagg_pd() || options_.enable_kvcache_store() || options_.enable_host_offload()) {
```
这样 host offload 场景也能正确包裹 `ConcurrentBlockManagerImpl`。

xllm/core/framework/model/model_input_params.h
```diff
@@ -629,16 +631,19 @@ struct AttentionInput {
 };
 
 enum class TransferType : uint8_t {
-  G2H = 0,  // global memory(KVCache store) to host memory(DRAM)
-  H2D = 1,  // host memory(DRAM) to device memory(HBM)
-  D2G = 2,  // host memory(DRAM) to global memory(KVCache store)
-  G2D = 3   // global memory(KVCache store) to device memory(HBM)
+  G2H = 0,    // global memory(KVCache store) to host memory(DRAM)
+  H2D = 1,    // host memory(DRAM) to device memory(HBM)
+  D2H2G = 2,  // device memory(HBM) to host memory(DRAM) to global
```
zhang-minchao: 这里新增了 `D2H2G = 2` 并把原来的 `D2G` 从 2 改成了 3，但 `proto/worker.proto` 里的 `TransferType` 枚举没有同步更新（proto 里仍然是 `D2G = 2`，且没有 `D2H2G` 条目）。

如果存在滚动升级场景（新旧 binary 混合通信），wire 上的值 2 在新 binary 被解释为 `D2H2G`，在旧 binary 被解释为 `D2G`，会导致 transfer 路由到错误的路径。

建议同步更新 `proto/worker.proto`：
```protobuf
enum TransferType {
  G2H = 0;
  H2D = 1;
  D2H2G = 2;
  D2G = 3;
  G2D = 4;
}
```

Kang-Meng: 顺序已经调整，D2H2G更新到proto中了（暂时没有用到），后续mooncake store的代码会用到

yingxudeng (dismissed):

yingxudeng (approved):

weizhehuang0827 (approved):

PR #1910, opened 2026-07-08
feat:  support  Wan2.2 distill model.

ethan686:


zhangjieshao98:
“Distilled model, steps=4, SP=4, guidance=1.0, CFG disabled: accuracy 0.972 compared with ligthx2v distill model, performance 26.43s.”

PR #2008, opened 2026-07-22
feat: support DSV4 device prefix cache (SWA + C4 + C128).

Kang-Meng:
# feat: support DSV4 device prefix cache (SWA + C4 + C128)

**Commit:** `d54b1af` · **Branch:** `feat_dsv4_prefixcache` · **Repo:** `xllm`
**Diff:** 30 files, +1242 / −576

## Background

DeepSeek-V4-Flash (DSV4) previously **could not use** the device prefix cache.
`master.cpp` carried a hard guard that, on detecting a `deepseek_v4` model type,
force-set `enable_prefix_cache=false` and logged a warning. This PR removes that
guard and completes `CompositeBlockManager` + prefix-cache support for DSV4's
three-segment KV layout (**SWA + C4 + C128**), so DSV4 can actually hit the
device-side prefix cache.

## What changed

1. **Removed the DSV4 fallback guard** (`master.cpp`, −11 lines).
   DSV4 now takes the normal composite prefix-cache path instead of being
   silently disabled.

2. **`LeafCombination` classification in `CompositeBlockManager`**
   (largest change; `composite_block_manager.cpp` +630). The leaf shape is
   classified once at construction and cached; the sequence-level orchestrators
   dispatch on it to pick the trim/mount strategy:
   - `FLAT_KV` — plain KV (normal / Qwen), no trim
   - `FLAT_KV_LINEAR` — Qwen3.5-Next GDN (KV + LINEAR restore)
   - `SWA_COMPRESSED` — **DSV4 (SWA + C4 + C128)**: cross-leaf min,
     C128-stride clamp, SWA tail-continuity, exact-repeat pop
   - `UNSUPPORTED` — prefix cache off / xtensor

   Adds `cache_full_blocks_for_sequence` to incrementally insert each leaf's
   newly-forwarded full blocks into its own prefix cache.

3. **Per-stride block-hash chains** (`sequence.h/.cpp`).
   DSV4 admission probes SWA/C4/C128 back-to-back each tick with different
   strides (base / 4×base / 128×base). Each stride now keeps its own
   incrementally-extended hash chain, so switching strides no longer discards
   and rebuilds the whole chain. Backed by `std::map` for pointer stability, so
   a `Slice` from `block_hashes()` survives inserts of other strides.

4. **`LinearStatePrefixCache` made gap-tolerant**
   (`linear_state_prefix_cache.cpp` +229). SWA slides out middle window blocks
   and LINEAR deeper hits subsume earlier ones, so `match()`/`insert()` tolerate
   middle misses (placeholders): the compute path walks the chain from token 0
   rather than seeding from the previous block, placeholders advance the cursor
   without being stamped/emplaced, and trailing misses are trimmed. The base
   `PrefixCache` remains solid-prefix.

5. **Interface cleanup**: dropped the `size_t* matched_tokens` out-param from
   `match()` (reach now expressed as `returned.size() * block_size_`); added an
   LRU-neutral `contains()`.

image

### DIFF
--- a/xllm/core/framework/prefix_cache/prefix_cache.h
+++ b/xllm/core/framework/prefix_cache/prefix_cache.h
@@ -41,17 +41,18 @@ inline size_t round_down(size_t n, size_t multiple) {
   return (n / multiple) * multiple;
 }
 
+// LRU-evicted prefix cache keyed by chained per-block hash. Solid-prefix
+// match by default; LinearStatePrefixCache overrides match with a
+// gap-tolerant walk for SWA / LINEAR leaves.
 class PrefixCache {
  public:
   struct Options {
     PROPERTY(int32_t, block_size) = 128;
     PROPERTY(BlockHasherType, hasher_type) = BlockHasherType::TEXT;
-    // Selects the concrete cache in create_prefix_cache: LINEAR builds the
-    // linear-state checkpoint index, everything else the base cache. A prefix
-    // cache indexes on a single block-boundary size (block_size above): for KV
-    // that is the KV block size, for the LINEAR index it is the checkpoint
-    // stride (one prefill chunk in tokens), routed into the same slot by
-    // BlockManagerImpl's constructor.
+    // Picks the concrete cache in create_prefix_cache: LINEAR / SWA build
+    // LinearStatePrefixCache (gap-tolerant), everything else this base class.
+    // A prefix cache indexes on a single stride (block_size above): KV block
+    // size for KV, prefill chunk stride for LINEAR, SWA base block for SWA.
     PROPERTY(BlockType, block_type) = BlockType::KV;
   };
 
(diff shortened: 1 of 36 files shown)


xllm/core/framework/prefix_cache/prefix_cache.h
```diff
@@ -161,73 +148,54 @@ class PrefixCache {
...
       node->next = lst_front.next;
       lst_front.next->prev = node;
-
       node->prev = &lst_front;
       lst_front.next = node;
     }
 
     Node* get_first() { return lst_front.next; }
 
-    // pop out node to the back of the LRU list
```
yingxudeng, line 187: 为什么移除了，原来的许多注释

yingxudeng (dismissed):

yingxudeng (dismissed):

yingxudeng (approved):

DongheJin (approved):

PR #2014, opened 2026-07-23
feat(lora): multi-tenant LoRA framework + attention/MLP wire-up (1/2)

cchh05:
See commit message for full details.

This is **PR 1 of 2** landing multi-tenant LoRA serving on xllm.

## This PR (~28 files, ~4300 lines)

- `xllm/core/framework/lora/` — new subsystem: runtime, registry, loader, context, metrics, config
- `xllm/core/layers/common/lora/` — 3 composition-based Linear wrappers (QKV/Column/Row parallel)
- `xllm/core/framework/model/model_input_params.h` — adds `adapter_ids` + `adapter_ids_per_token` fields
- `xllm/models/llm/llm_model_base.h` — pushes LoRAContext frame in forward
- `xllm/core/layers/common/qwen2_attention.{h,cpp}`, `dense_mlp.{h,cpp}` — swaps Linear members to LoRA-wrapped drop-ins

## PR 2/2 will follow with

- Request path adapter_id propagation (RequestParams → Sequence → BatchInputBuilder)
- HTTP endpoints `/v1/load_lora_adapter`, `/v1/unload_lora_adapter`, `/v1/lora_adapters`, `/v1/lora_stats`
- Chat routing (adapter name-as-model + explicit `lora_name` field)

## Not in either PR (separate CLs)

- Qwen3-Next hybrid attention (Qwen3.5-122B) support
- Fused MoE expert LoRA delta injection

## Validation

End-to-end verified on the pre-split fork branch:
- Qwen3-30B-A3B-Instruct-2507 TP=8 with a real Megatron-trained PEFT LoRA (r=16, target={q,k,v,o}_proj): **+9pp compliance-rate and +0.04 gt-Jaccard vs base** on 200-sample business eval; base output pixel-perfect unchanged (fix is non-invasive).
- 30min sustained load stress: **5363 req, err=0.00%, HBM drift +3MB**.
- Divergence sweep N=50 under temperature=0: **70% divergence rate** (systematic delta, not noise).

PR #2015, opened 2026-07-23
feat(lora): request path adapter_id propagation + chat routing (2/2)

cchh05:
See commit message for full plumbing chain diagram.

This is **PR 2 of 2** landing multi-tenant LoRA serving on xllm.
Depends on #2014.

## This PR (10 files, ~60 lines)

Wires the LoRA `adapter_id` from `ChatServiceImpl` all the way to
`ModelInputParams.adapter_ids` so the LoRA-wrapped Linear layers landed
in PR #2014 can pick up per-sequence routing at forward time.

**Request/batch path:**
- `request_params.h`, `request_state.h`, `sequence.h`/`.cpp`, `request.cpp` — carry `adapter_id`
- `batch_input_builder.h`/`.cpp` — collect per-seq `adapter_ids` into `ModelInputParams`
- `forward_params.h` — adds `ForwardInput::adapter_ids`

**Distributed runtime:**
- `llm_master.cpp` — `RequestParams::adapter_id` → `RequestState::adapter_id`

**Chat routing:**
- `chat_service_impl.cpp` — `model` field looked up in `LoRARegistry` via `lookup_and_pin`; if it resolves to a registered adapter, populate `request_params.adapter_id`.

## Not in this PR (follow-up CLs)

- HTTP endpoints `/v1/load_lora_adapter`, `/v1/unload_lora_adapter`, `/v1/lora_adapters`, `/v1/lora_stats` — requires new `master->load_lora_broadcast()` / `unload_lora_broadcast()` engine APIs
- Non-rank0 broadcast of `adapter_ids` via shm channel — for TP=1 this PR is already end-to-end usable; TP>1 will silent-fallback to base on non-rank0 workers until broadcast lands
- Full drain lifecycle: unpin adapter on chat request finish

## Rollout

Depends on #2014 landing first. Static adapter preload via `--lora_modules=<name>=<path>` (existing gflag from #2014) works end-to-end with this CL on TP=1; dynamic hot-load and TP>1 broadcast follow in later CLs.

## Fork validation

End-to-end verified on the pre-split fork branch (see #2014 body for numbers).


cchh05:
Note on CI failures — this is an upstream CI infrastructure issue, not a code problem.

All 4 `check-sensitive` failures on `arm_64 NPU` failed at the **checkout stage**, with identical error:

```
fatal: unable to access 'https://gh-proxy.test.osinfra.cn/https://github.com/xLLM-AI/xllm/':
The requested URL returned error: 502
```

The CI runner uses `gh-proxy.test.osinfra.cn` as a GitHub reverse proxy, and that proxy has been returning 502 Bad Gateway consistently across attempts 1-4. The build has never actually run on our diff.

PR #2014 (dependency) is experiencing the exact same failure mode.

Not planning to push new commits — will wait for the upstream proxy to recover, then request a re-run. If someone from the infra team can help investigate the proxy or bypass it (direct GitHub fetch), that would unblock both PRs.

DongheJin:
Commit messages must conform to the specifications and must end with a dot.

cchh05:
Thanks for the review! Amended both commit message headlines (PR 2/2 and the ifdef-fix commit) to end with a period. Also rebased onto the amended PR #2014 to keep the branch clean. Diff-only fix — code is byte-identical (`git diff 33b72ab8 a52b6923 --stat` is empty). Force-pushed. Please take another look when convenient.

cchh05:
### Rebase update — 2026-08-06

Fresh rebase onto the updated PR #2014 tip (which now includes the framework runtime-init / SHM adapter_ids / is_weight_loaded fixes). Same content as the previous head of this PR plus one small follow-up commit:

`fix(lora): route registered LoRA names through base master in get_model_master` — the existing PR already stamps `request_params.adapter_id` when the model name matches a registered adapter, but `get_model_master` returned `nullptr` for the LoRA name so the request short-circuited to `Model not supported` before reaching the adapter-aware forward path. Falling back to `master_` when LoRARuntime is enabled and the name resolves to a registered adapter closes the loop.

Validated as part of the same 20/20 sample run on Qwen3.5-122B-A10B TP=16 + biz122b adapter — see PR #2014 for the numbers.

@HuHeng dependencies: this PR now sits on top of the updated #2014 (framework fixes). Please review after #2014 lands.

PR #2016, opened 2026-07-23
bugfix: reconcile max_seqs_per_batch with max_concurrent_requests.

Kang-Meng:
- Align the scheduler batch cap and the service admission cap at
  startup in initialize_configs(): min when both > 0, the non-zero
  one when one is 0, LOG(FATAL) when both are 0. Downstream users
  (Master, Engine, BlockManagerPool, RateLimiter, ProfileManager)
  now see the same effective cap on both fields.
- Drop the scheduler default from 1024 to 200 so it matches
  ServiceConfig::max_concurrent_requests out of the box (keeps the
  gflag DEFINE_* and the PROPERTY in-class default in sync per
  custom-code-style §10).
- Fix VLM and Rec engines silently dropping max_seqs_per_batch
  when building BlockManagerPool::Options -- the SINGLE block pool
  was falling back to a 2-slot capacity (1 padding + 1 usable),
  effectively capping concurrent sequences at 1 on those paths.
- Tighten a few unit tests that relied on the pool default being
  0 by setting .max_seqs_per_batch(0) explicitly, so the intent to
  drive the SINGLE pool from num_single_blocks alone stays local
  and does not depend on Options' default value.

image

### DIFF
--- a/tests/core/framework/block/block_manager_test.cpp
+++ b/tests/core/framework/block/block_manager_test.cpp
@@ -24,7 +24,6 @@ limitations under the License.
 #include "block_manager_pool.h"
 #include "common/global_flags.h"
 #include "core/framework/config/scheduler_config.h"
-#include "core/framework/config/service_config.h"
 #include "framework/block/block_manager_pool_test_peer.h"
 #include "framework/block/linear_state_block_manager.h"
 #include "framework/model/model_input_params.h"
@@ -164,8 +163,12 @@ BlockManagerPool::Options make_linear_state_pool_options(
   BlockManagerPool::Options options;
   options.num_blocks(8).host_num_blocks(0).block_size(4).enable_prefix_cache(
       true);
-  options.num_single_blocks(4).enable_linear_state(true).linear_state_num_slots(
-      linear_state_num_slots);
+  // Zero out max_seqs_per_batch so the SINGLE pool is sized purely by
+  // num_single_blocks (otherwise the pool takes the max of the two).
+  options.max_seqs_per_batch(0)
+      .num_single_blocks(4)
+      .enable_linear_state(true)
+      .linear_state_num_slots(linear_state_num_slots);
   return options;
 }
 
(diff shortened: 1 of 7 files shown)


weizhehuang0827 (dismissed):

tests/core/framework/block/block_manager_test.cpp
```diff
@@ -375,20 +377,21 @@ TEST(BlockManagerPoolTest, TryAllocateKvFailureRollsBackSingleBlock) {
...
+  // 8+2 path would win; force num_single_blocks to be the winner by lowering
+  // the scheduler side).
   options.num_blocks(16)
       .host_num_blocks(0)
       .block_size(1)
+      .max_seqs_per_batch(0)
       .num_single_blocks(4)
       .enable_prefix_cache(false);
   options.enable_linear_state(true).linear_state_num_slots(64);
+  options.enable_linear_state(true).linear_state_num_slots(64);
```
yingxudeng: 重复两行

xllm/xllm.cpp
```diff
@@ -84,6 +85,33 @@ void initialize_configs() {
...
+    const int32_t service_cap = service_config.max_concurrent_requests();
+    int32_t effective_cap = 0;
+    if (scheduler_cap > 0 && service_cap > 0) {
+      effective_cap = std::min(scheduler_cap, service_cap);
+    } else if (scheduler_cap > 0) {
+      effective_cap = scheduler_cap;
+    } else if (service_cap > 0) {
+      effective_cap = service_cap;
+    } else {
+      LOG(FATAL) << "Both max_seqs_per_batch and max_concurrent_requests are "
```
yingxudeng, line 109: max_concurrent_requests=0 原来这样代表无限制，现在会变成未设置，并且覆盖成 200，这是语意的变化是否符合预期

yingxudeng:
目前 cicd npu 失败，请补充本地 npu 通过所有单测的截图

yingxudeng (approved):

weizhehuang0827 (approved):

PR #2092, opened 2026-07-31
feat: add pd prefill short request first scheduling.

immengzi:
## Description

Adds opt-in `ShortRequestFirst` scheduling for PD-prefill requests.

- Adds `enable_short_request_first`, `short_request_first_threshold`, and `short_request_first_long_max_wait_ms` through CLI flags and JSON configuration.
- Splits online PD-prefill requests into immediate, short, and long queues. Requests with partial KV state or preemption recovery are scheduled first; fresh short requests are then prioritized over fresh long requests.
- Supports optional long-request aging to prevent starvation. A value of `0` retains strict short-request-first behavior.
- Keeps offline requests in the existing queue and preserves online-request scheduling priority.
- Restricts the feature to FCFS PD-prefill/MIX scheduling with chunked prefill enabled.
- Adds queue-depth and aged-long-promotion metrics.
- Adds configuration, queue, and scheduler unit tests, together with PD startup and AISBench benchmark/replay helper scripts.

## Related Issues

N/A

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Performance improvement
- [ ] Refactor
- [ ] Documentation
- [x] Test
- [ ] Build or CI

### Pre-commit Checks

### Self Review

### Build and Test Coverage

## Reviewer Notes

No known functional issues.


xllm/core/scheduler/disagg_pd_chunked_prefill_scheduler.cpp
weizhehuang0827, line 1: The scheduler has been refactored in the main branch, and `disagg_pd_chunked_prefill_scheduler` has been removed. Additionally, the strategy in this PR appears to be a subset of the `priority_strategy=multi_slo_and_priority` case; you can check that out.

xllm/core/scheduler/short_request_first_policy.cpp
```diff
@@ -0,0 +1,125 @@
...
+#include "scheduler/scheduler_policy.h"
+
+namespace xllm {
+namespace {
+
+// Sort rank for ShortRequestFirst ordering; lower ranks are scheduled first.
+int32_t short_request_first_rank(ShortRequestFirstRequestClass request_class,
+                                 const std::shared_ptr<Request>& request,
+                                 const std::shared_ptr<Request>& promoted) {
+  if (request_class == ShortRequestFirstRequestClass::IMMEDIATE) {
```
weizhehuang0827: Is IMMEDIATE the same as a chunked request? We already have a chunked queue.

immengzi: Good catch. In the current architecture, chunked prefill continuations that hold partial KV are routed to the separate chunk_queue and scheduled ahead of the prefill queue, so they never enter the queue ShortRequestFirst sorts — that part of the old IMMEDIATE logic was redundant and I've removed it. IMMEDIATE now only covers preempted requests: the generic scheduler already pushes preempted requests to the front of the prefill queue once their KV is deallocated, and ShortRequestFirst keeps that front-of-queue priority explicit in the sort (preempted → rank 0). So it doesn't overlap with the chunk queue.

xllm/core/scheduler/disagg_pd_chunked_prefill_scheduler.cpp
immengzi: Thanks for the suggestion. I've reworked the implementation into a policy/comparator-based approach instead of a separate queue, following the current scheduler-policy structure. The updated branch is pushed — please take another look when you get a chance.

immengzi: Regarding multi_slo_and_prio: this change is a complementary TTFT-oriented waiting-order optimization, not a subset.

xllm/core/scheduler/continuous_scheduler.cpp
```diff
@@ -187,6 +208,14 @@ bool ContinuousScheduler::add_request(std::shared_ptr<Request>& request) {
 }
 
 void ContinuousScheduler::create_queues(const Options& options) {
+  const SchedulerConfig& scheduler_config = SchedulerConfig::get_instance();
```
weizhehuang0827: move this check to scheduler_policy factory

xllm/core/scheduler/continuous_scheduler.cpp
```diff
@@ -47,6 +47,27 @@ limitations under the License.
 #include "util/utils.h"
 
 namespace xllm {
+namespace {
+
+void validate_short_request_first_options(
```
weizhehuang0827: also move this

xllm/core/scheduler/continuous_scheduler.cpp
immengzi: Done. Moved the validation and the startup log into create_scheduler_policy() in scheduler_policy.cpp (commit 2d666e21).

xllm/core/scheduler/continuous_scheduler.cpp
immengzi: Done. Moved the validation and the startup log into create_scheduler_policy() in scheduler_policy.cpp (commit 2d666e21).

xllm/core/scheduler/continuous_scheduler.cpp
```diff
@@ -201,6 +201,35 @@ void ContinuousScheduler::create_queues(const Options& options) {
   }
 }
 
+void ContinuousScheduler::report_short_request_first_metrics() {
```
weizhehuang0827: move to your specific policy

xllm/core/scheduler/short_request_first_policy.cpp
```diff
@@ -0,0 +1,128 @@
...
+                                                           int32_t threshold) {
+  CHECK(!request.sequences().empty());
+  Sequence* sequence = request.sequences()[0].get();
+  CHECK(sequence != nullptr);
+
+  // Chunked prefill continuations that hold KV blocks are routed to the
+  // chunk queue and scheduled ahead of the prefill queue, so the prefill
+  // queue only contains fresh requests and preempted requests (whose KV has
+  // already been deallocated). IMMEDIATE therefore means "preempted".
+  if (request.preempted()) {
```
weizhehuang0827: Requests on a prefill instance will never be preempted; only decode requests will preempt requests. The `IMMEDIATE` is redundant.

xllm/core/scheduler/continuous_scheduler.cpp
immengzi: Done. Removed report_short_request_first_metrics() from ContinuousScheduler; the ShortRequestFirst lane gauges are now reported by ShortRequestFirstPolicy::report_metrics() (commit f20b1e47).

xllm/core/scheduler/short_request_first_policy.cpp
immengzi: Agreed. Preemption only happens on the decode side, so requests on the PD-prefill instance ShortRequestFirst targets never carry the preempted flag — the IMMEDIATE lane was unreachable. Removed it along with its gauge and tests (commit 3defc102).

xllm/core/scheduler/short_request_first_policy.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+  std::shared_ptr<Request> promoted;
+  if (long_max_wait_ms > 0.0 && has_short && has_long &&
+      oldest_long != nullptr &&
+      now - oldest_long->created_time() >=
+          absl::Milliseconds(long_max_wait_ms)) {
+    promoted = oldest_long;
+  }
+
+  queue.sort([threshold, promoted](const std::shared_ptr<Request>& a,
```
weizhehuang0827: You need to set your own priority_strategy and put the request order func in /request/priority_comparator.cpp

immengzi: Thanks. ShortRequestFirst now has its own priority_strategy (short_request_first) instead of reusing fcfs, and the ordering logic lives in framework/request/priority_comparator.{h,cpp}:
ShortRequestFirstComparator ranks the promoted LONG head, then SHORT, then the remaining LONG requests, with ties broken by created_time;
since a pairwise comparator can't see the queue head, a small range-based helper in the same file selects the single aged LONG head that may bypass SHORT requests;
ShortRequestFirstPolicy scans the queue once and sorts with the comparator.
Validation now requires priority_strategy=short_request_first (commits 0c843abd, 441f39ed, 537be114).

xllm/core/framework/config/scheduler_config.cpp
```diff
@@ -56,7 +58,21 @@ DEFINE_int32(
 
 DEFINE_string(priority_strategy,
               "fcfs",
-              "Priority strategy for requests(e.g. fcfs, priority, deadline).");
+              "Priority strategy for requests(e.g. fcfs, priority, deadline, "
+              "short_request_first).");
+
+DEFINE_bool(enable_short_request_first,
```
weizhehuang0827: remove

immengzi: Thanks for the feedback. Removed enable_short_request_first — it was redundant with priority_strategy=short_request_first, which the factory already required. SRF is now activated solely via --priority_strategy=short_request_first, with the config property, CLI/JSON mappings, and related tests cleaned up accordingly. Also dropped the now-unreachable strategy re-check and added an explicit error for enable_mix_batch=true + short_request_first instead of silently falling back to FCFS.

weizhehuang0827 (dismissed):

PR #2254, opened 2026-08-17
bugfix: do not build a tokenizer for draft-engine workers.

weizhehuang0827:
## Description

`WorkerImpl::init_model` unconditionally built a tokenizer and asserted it was
non-null, and after init every worker also warmed the JSON-object grammar
(which hard-requires a tokenizer):

```cpp
auto tokenizer = model_loader->tokenizer();
CHECK(tokenizer != nullptr);
...
ensure_json_object_grammar(/*reasoning_enabled=*/false);  // CHECK(tokenizer_ != nullptr)
```

A draft engine is fed token ids and detokenized by the target, so its weights
directory may ship **no** tokenizer files — the DFlash / DSpark draft
checkpoints contain only `config.json` + weights. In that case
`TokenizerFactory` finds no tokenizer args, falls back to the SentencePiece
loader, fails to open a nonexistent `tokenizer.model`, and `LOG(FATAL)` aborts
the worker during init. Even with the tokenizer build skipped, the JSON-grammar
warmup then aborts on the null tokenizer. Either way any multi-process
(worker-server) DFlash / DSpark launch crashes on startup.

This guards both draft-engine paths:
1. Skip building the tokenizer for draft engines (matching the existing
   `if (!is_draft_engine())` guard already used at the engine level).
2. Skip the post-init JSON-grammar warmup for draft engines.

## Why it is safe

The worker tokenizer is used only by the **target** worker:
- vocab-size reconciliation (already inside `if (!is_draft_engine())`), and
- JSON-object grammar (warmup + `restore_json_object_states` /
  `ensure_json_object_grammar`), which only runs when a request carries JSON
  grammar state — a target-only path.

A draft worker keeps `tokenizer_` null and never reaches those uses.

## Test

- DFlash / DSpark multi-process launch (`nnodes >= 2`) previously aborted at
  init — first with `Failed to load SentencePiece model from
  .../tokenizer.model`, then (tokenizer skipped) with `JSON object grammar
  requires a worker tokenizer`. With this change the draft worker initializes
  without a tokenizer and the server comes up.

### DIFF
--- a/xllm/core/runtime/worker_impl.cpp
+++ b/xllm/core/runtime/worker_impl.cpp
@@ -1871,15 +1871,6 @@ bool WorkerImpl::init_model(const std::string& model_weights_path,
 
   auto model_loader = ModelLoader::create(model_weights_path);
   model_weights_path_ = std::move(model_weights_path);
-  // A draft engine is fed token ids and detokenized by the target, so it owns
-  // no tokenizer (its weights dir may not even ship tokenizer files, e.g. the
-  // DFlash / DSpark draft checkpoints). Only the target worker builds one; its
-  // JSON grammar restoration below is target-only and never runs on a draft.
-  std::unique_ptr<Tokenizer> tokenizer;
-  if (!options_.is_draft_engine()) {
-    tokenizer = model_loader->tokenizer();
-    CHECK(tokenizer != nullptr);
-  }
 
   auto args = model_loader->model_args();
   auto quant_args = model_loader->quant_args();
@@ -1887,9 +1878,16 @@ bool WorkerImpl::init_model(const std::string& model_weights_path,
   args.embedding_mode(embedding_mode);
   torch::ScalarType dtype = util::parse_dtype(args.dtype(), device_);
 
-  // Only the target engine reconciles tokenizer and model vocab sizes. Draft
-  // engines do not detokenize output and own no tokenizer.
+  // A draft engine is fed token ids and detokenized by the target, so it owns
+  // no tokenizer (its weights dir may not even ship tokenizer files, e.g. the
+  // DFlash / DSpark draft checkpoints). Only the target worker builds one and
+  // reconciles vocab sizes; its JSON grammar restoration below is target-only
+  // and never runs on a draft.
+  std::unique_ptr<Tokenizer> tokenizer;
   if (!options_.is_draft_engine()) {
+    tokenizer = model_loader->tokenizer();
+    CHECK(tokenizer != nullptr);
+
     const int64_t tokenizer_vocab_size = tokenizer->vocab_size();
     int64_t model_vocab_size = args.vocab_size();
     // use tokenizer vocab size if model vocab size is not set
--- a/xllm/core/runtime/worker_impl.cpp
+++ b/xllm/core/runtime/worker_impl.cpp
@@ -1871,8 +1871,15 @@ bool WorkerImpl::init_model(const std::string& model_weights_path,
 
   auto model_loader = ModelLoader::create(model_weights_path);
   model_weights_path_ = std::move(model_weights_path);
-  auto tokenizer = model_loader->tokenizer();
-  CHECK(tokenizer != nullptr);
+  // A draft engine is fed token ids and detokenized by the target, so it owns
+  // no tokenizer (its weights dir may not even ship tokenizer files, e.g. the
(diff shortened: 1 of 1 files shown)


yinjiawei01 (dismissed):

xllm/core/runtime/worker_impl.cpp
```diff
@@ -1871,8 +1871,15 @@ bool WorkerImpl::init_model(const std::string& model_weights_path,
...
   auto model_loader = ModelLoader::create(model_weights_path);
   model_weights_path_ = std::move(model_weights_path);
-  auto tokenizer = model_loader->tokenizer();
-  CHECK(tokenizer != nullptr);
+  // A draft engine is fed token ids and detokenized by the target, so it owns
+  // no tokenizer (its weights dir may not even ship tokenizer files, e.g. the
+  // DFlash / DSpark draft checkpoints). Only the target worker builds one; its
+  // JSON grammar restoration below is target-only and never runs on a draft.
+  std::unique_ptr<Tokenizer> tokenizer;
+  if (!options_.is_draft_engine()) {
```
ustcfy: This turns "draft engine has no tokenizer" from *incidentally harmless* into a *hard invariant*: after this change a draft worker's `tokenizer_` is always null, so any forward input that reaches this worker carrying a non-empty `json_object_state_snapshots` would now abort in `restore_json_object_states` (`CHECK(tokenizer_ != nullptr)`, worker_impl.cpp:1300) rather than merely skipping a warmup.

Code tracing supports the safety claim — the draft input builder (`speculative_worker_impl.cpp` / specBuilder) only writes token/position/sampling tensors and never populates the JSON grammar fields, and `dflash_worker_impl.cpp` has no `json_object` references. But the forward SHM (de)serialization path is shared by all workers, so whether snapshots ever reach the draft process is ultimately decided by target-side dispatch and can't be fully ruled out statically.

xllm/core/runtime/worker_impl.cpp
```diff
@@ -1881,8 +1888,7 @@ bool WorkerImpl::init_model(const std::string& model_weights_path,
   torch::ScalarType dtype = util::parse_dtype(args.dtype(), device_);
 
   // Only the target engine reconciles tokenizer and model vocab sizes. Draft
-  // engines do not detokenize output, but worker-local JSON grammar state
-  // restoration still requires retaining the tokenizer below.
+  // engines do not detokenize output and own no tokenizer.
   if (!options_.is_draft_engine()) {
```
ustcfy, line 1887: Optional / non-blocking (feel free to ignore for a focused bugfix): this is now the second `if (!options_.is_draft_engine())` in a row, separated only by the tokenizer-independent `args` / `quant_args` / `dtype` setup. Since that setup has no dependency on the tokenizer, hoisting it above the guard lets the two blocks collapse into one — dropping a redundant condition and a duplicated comment:

```cpp
auto args = model_loader->model_args();
auto quant_args = model_loader->quant_args();
const bool embedding_mode = options_.task_type() == "embed";
args.embedding_mode(embedding_mode);
torch::ScalarType dtype = util::parse_dtype(args.dtype(), device_);

// A draft engine is fed token ids and detokenized by the target, so it owns
// no tokenizer (its weights dir may not even ship tokenizer files, e.g. the
// DFlash / DSpark draft checkpoints). Only the target worker builds one and
// reconciles vocab sizes; its JSON grammar restoration below is target-only.
std::unique_ptr<Tokenizer> tokenizer;
if (!options_.is_draft_engine()) {
  tokenizer = model_loader->tokenizer();
  CHECK(tokenizer != nullptr);

  const int64_t tokenizer_vocab_size = tokenizer->vocab_size();
  int64_t model_vocab_size = args.vocab_size();
  if (model_vocab_size <= 0) {
    LOG(WARNING) << "Model vocab size is not set, using tokenizer vocab size: "
                 << tokenizer_vocab_size;
    args.vocab_size(tokenizer_vocab_size);
  } else if (tokenizer_vocab_size > model_vocab_size) {
    LOG(WARNING) << "Unsafe vocab mismatch: tokenizer: "
                 << tokenizer_vocab_size << ", model: " << model_vocab_size;
  }
}
tokenizer_ = std::move(tokenizer);
```

The reorder is behavior-preserving (`args`/`quant_args`/`dtype` don't touch the tokenizer). Purely a readability nit — the current form is fine too.

ustcfy (approved):

yinjiawei01 (approved):

PR #2255, opened 2026-08-17
bugfix: enable speculative decode under dp4 when some DP ranks are idle.

ShuweiShen772:
## Problem
Under dp4_ep4 with MTP (speculative decode) and enable_graph=False, dp4+MTP is
*slower* than dp4 without MTP (36.1ms vs 32ms TPOT). Profiling shows the validate
stage never runs (validation latency = 0, draft_tokens = 0): speculation silently
falls back to the plain decode path while still paying the draft cost.

## Root cause
`should_run_speculative_decode()` requires **all** DP ranks to be in decode
(`dp_is_decode` all ones). With bs < dp_size, idle DP ranks carry no tokens and keep
`dp_is_decode = 0`. The backfill in `llm_engine.cpp` that marks empty ranks as decode
is gated on `enable_graph=True`, so under `enable_graph=False` those idle ranks veto
speculation for the whole step.

## Fix
Only ranks that actually carry tokens (`dp_global_token_nums != 0`) gate the decision;
idle ranks no longer veto. dp1 and the non-speculative fallback are unaffected.

## Validation (16-card Ascend, V3-server103 w8a8, dp4_ep4, enable_graph=False)
| metric | before | after |
|---|---|---|
| avg TPOT | 36.1 ms | **16.0 ms** |
| validate latency | 0 | 0.194 s |
| draft tokens | 0 | 7696 (accept 83%) |

### DIFF
--- a/xllm/core/runtime/speculative_worker_impl.cpp
+++ b/xllm/core/runtime/speculative_worker_impl.cpp
@@ -104,9 +104,24 @@ bool should_run_speculative_decode(const ModelInputParams& params) {
     return false;
   }
 
-  return std::all_of(dp_is_decode.begin(),
-                     dp_is_decode.end(),
-                     [](int32_t is_decode) { return is_decode == 1; });
+  // Idle DP ranks (no scheduled tokens this step) must not veto speculative
+  // decode for the active ranks. Under enable_graph=False these idle ranks keep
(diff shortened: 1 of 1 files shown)


Kang-Meng (dismissed):
LGTM

Kang-Meng (approved):
LGTM

yingxudeng (approved):

PR #2298, opened 2026-09-02
refactor: record inference rpc metrics when requests complete.

liutongxuan:
Replace xllm_metrics.h with RpcRequestMetrics owned by Call so ok/fail counters fire at request end instead of ClosureGuard::release().

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Performance improvement
- [ ] Refactor
- [ ] Documentation
- [ ] Test
- [ ] Build or CI

## Pull Request Checklist

- [ ] The PR title and each commit message follow the xLLM commit format: `<type>: <subject>`.

### Pre-commit Checks

- [ ] I have installed `pre-commit` by running `pip install pre-commit` or an equivalent command.
- [ ] I have installed the hooks with `pre-commit install`.
- [ ] I have run `pre-commit run --all-files` and fixed any reported issues.

### Self Review

- [ ] I have self-reviewed the code according to `.agents/skills/code-review/references/custom-code-style.md`, especially code written or assisted by AI.

### Build and Test Coverage


xllm/api_service/call.h
```diff
@@ -55,6 +58,7 @@ class Call {
 
  protected:
   brpc::Controller* controller_;
+  RpcRequestMetrics rpc_metrics_;
```
zhang-minchao, line 72: 待确认：`RpcRequestMetrics` 现在作为 `Call` 的成员保存裸 `brpc::Controller*`。它的析构函数定义在 `xllm/api_service/rpc_request_metrics.cpp:29`：

```cpp
RpcRequestMetrics::~RpcRequestMetrics() { record_out(); }
```

而 `record_out()` 在同文件 `:50-56` 解引用该指针：

```cpp
if (!controller_->Failed()) {
    ...
}
```

但 `controller_` 的释放时机似乎可能早于 `rpc_metrics_`：

- `NonStreamCall::~NonStreamCall()` 先调用 `done_->Run()`；
- 非流式 `StreamCall::~StreamCall()` 也先调用 `done_->Run()`；
- 流式 `StreamCall` 则在构造函数中调用 `done_->Run()`。

如果 brpc 的 `done_->Run()` 在返回前释放 controller，那么随后执行的 `RpcRequestMetrics::~RpcRequestMetrics()` 就会调用 `record_out()`，进而访问已经释放的 `controller_->Failed()`，可能发生 UAF。请确认是否应在 `done_->Run()` 前记录 metrics，或以其他方式保证 controller 在 metrics 记录完成前仍存活。

**brpc 源码依据**

PR 使用的 brpc gitlink 是 `67d6372a`；xLLM 从 `415f5378` 更新到该版本时，以下 ownership 代码未变化。

HTTP response sender 直接持有 controller：

`src/brpc/policy/http_rpc_protocol.cpp:738`
```cpp
std::unique_ptr<Controller, LogErrorTextAndDelete> _cntl;
```

HTTP completion callback 先执行回调，再销毁 sender：

`src/brpc/policy/http_rpc_protocol.cpp:748-752`
```cpp
_sender._cntl->CallAfterRpcResp(...);
delete this;
```

sender 销毁后，deleter 会删除 controller：

`src/brpc/protocol.cpp:257-272`
```cpp
if (_delete_cntl) {
    delete c;
}
```

brpc 代码也明确说明 HTTP 的 `cntl`、`req`、`res` 会在 `done` 内删除（`src/brpc/policy/http_rpc_protocol.cpp:1456-1457`）。

流式请求在构造函数中触发同一个 `done`：

`xllm/api_service/stream_call.h:58-69`
```cpp
stream_ = request_->stream();
if (stream_) {
    pa_ = controller_->CreateProgressiveAttachment();
    done_->Run();
}
```

`CreateProgressiveAttachment()` 只把 socket 引用交给 progressive attachment（`src/brpc/controller.cpp:1506-1514`），后续 chunk 通过该 socket 发送，并不会延长 controller 生命周期。因此上述释放顺序在流式 HTTP 中同样适用。

liutongxuan: done，已修复

xllm/api_service/rpc_request_metrics.h
```diff
@@ -0,0 +1,63 @@
...
+
+  void mark_failed(int32_t error_code = 0);
+
+ private:
+  void record_out();
+
+  const brpc::Controller* controller_ = nullptr;
+  bool active_ = false;
+  bool failed_ = false;
+  int32_t error_code_ = 0;
```
ustcfy, line 60: `failed_` and `error_code_` are written from callback threads (`chat_service_impl.cpp` / `completion_service_impl.cpp` capture `metrics` into a `std::function` the master invokes) and read on `record_out()`. Non-atomic writes are technically a data race. Please make them `std::atomic<bool>` / `std::atomic<int32_t>` — or document on the class that "callback invocations must be serialized relative to destruction," so a future master change to parallel notifications doesn't silently break this.

ustcfy (commented):
Follow-up review from an outside contributor — quality/simplification pass on top of the existing UAF-focused reviews. Skipping the two concerns already addressed (UAF fix in dtor ordering; mark_failed on `handle_rpc_response` returning false). Comments below are inline; the biggest lever is the `api_service_impl.h` one — pushing metric ownership up to APIService dissolves several smaller comments at once.

xllm/api_service/rpc_request_metrics.h
```diff
@@ -0,0 +1,63 @@
...
+  // use the snapshot.
+  void detach();
+
+  void mark_failed(int32_t error_code = 0);
+
+ private:
+  void record_out();
+
+  const brpc::Controller* controller_ = nullptr;
+  bool active_ = false;
```
ustcfy, line 58: `active_` is redundant with `!recorded_` — every state transition already implies whether we've recorded. Consider collapsing to a single `recorded_` flag (or the three-state `enum { LIVE, SNAPSHOTTED, RECORDED }`) — ctor sets `recorded_ = (controller == nullptr)` and `record_out()` sets it true. Removes one field and one thing to keep in sync.

xllm/api_service/rpc_request_metrics.h
```diff
@@ -0,0 +1,63 @@
...
+
+  RpcRequestMetrics(const RpcRequestMetrics&) = delete;
+  RpcRequestMetrics& operator=(const RpcRequestMetrics&) = delete;
+
+  RpcRequestMetrics(RpcRequestMetrics&& other) noexcept;
+  RpcRequestMetrics& operator=(RpcRequestMetrics&& other) noexcept;
+
+  // Snapshot Failed()/ErrorCode() and record immediately. Must be called
+  // before done->Run() recycles the controller. Idempotent.
+  void finish(const brpc::Controller* controller);
```
ustcfy, line 45: The `controller` parameter is redundant with the class-stored `controller_`; every production caller passes `controller_` back in through `Call::finish_rpc_metrics()`. Consider dropping the parameter and reading `this->controller_`; a test that wants to inject a different controller can pass it at construction instead. Also, `detach()`, `finish(nullptr)`, and `~()` all funnel to the same `record_out()` exit — collapsing them to one private `record()` + one public `finish()` reduces the API surface by a method.

xllm/api_service/rpc_request_metrics.h
```diff
@@ -0,0 +1,63 @@
...
+  explicit RpcRequestMetrics(brpc::Controller* controller);
+
+  ~RpcRequestMetrics();
+
+  RpcRequestMetrics(const RpcRequestMetrics&) = delete;
+  RpcRequestMetrics& operator=(const RpcRequestMetrics&) = delete;
+
+  RpcRequestMetrics(RpcRequestMetrics&& other) noexcept;
+  RpcRequestMetrics& operator=(RpcRequestMetrics&& other) noexcept;
+
```
ustcfy, line 42: Nit: "Idempotent" is slightly misleading — after `finish()`, a later `mark_failed()` still mutates `failed_` / `error_code_`, even though counter side-effects only fire once. Suggest "records at most once."

xllm/api_service/rpc_request_metrics.cpp
```diff
@@ -0,0 +1,98 @@
...
+  }
+  controller_ = nullptr;
+  record_out();
+}
+
+void RpcRequestMetrics::detach() { controller_ = nullptr; }
+
+void RpcRequestMetrics::mark_failed(int32_t error_code) {
+  failed_ = true;
+  error_code_ = error_code;
```
ustcfy, line 77: `mark_failed` unconditionally overwrites `error_code_`. If `finish()` has already captured `ELIMIT` from the controller, a later `mark_rpc_failed()` (default arg `0`) silently clobbers it. Counters are still correct because `active_` gates `record_out()`, but the residual `error_code_` becomes misleading. Prefer first-failure-wins:
```cpp
if (!failed_) { failed_ = true; error_code_ = error_code; }
```

xllm/api_service/call.h
```diff
@@ -17,9 +17,12 @@ limitations under the License.
 
 #include <brpc/controller.h>
```
ustcfy, line 18: Nit: `api_service/rpc_request_metrics.h` (included on line 24) already pulls in `<brpc/controller.h>`. Style §2 discourages duplicate includes; either drop this one or keep both with a "self-sufficient header" convention.

xllm/api_service/call.h
```diff
@@ -50,11 +54,22 @@ class Call {
 
   virtual bool is_disconnected() const = 0;
 
+  void mark_rpc_failed(int32_t error_code = 0);
```
ustcfy, line 57: `Call::mark_rpc_failed` is a pure forwarder to `rpc_metrics_.mark_failed()` with one caller (the test at `rpc_request_metrics_test.cpp`). Consider exposing a `rpc_metrics()` getter, or making the member `protected:` for derived Calls, rather than growing the base `Call` API just to forward. Similarly `release_controller()` is a single-purpose forwarder — both disappear if metric ownership moves up to APIService (see comment on `api_service_impl.h`).

xllm/api_service/api_service_impl.h
```diff
@@ -45,7 +45,8 @@ class APIServiceImpl {
 
   virtual void process_async_impl(std::shared_ptr<CallType> call) = 0;
 
-  virtual void process_async_rpc_impl(const RequestType* request) {
+  virtual void process_async_rpc_impl(const RequestType* request,
+                                      RpcRequestMetrics rpc_metrics = {}) {
```
ustcfy, line 49: The default `RpcRequestMetrics rpc_metrics = {}` propagates through `chat_service_impl.h` and `completion_service_impl.h` too. Any future caller that omits the arg silently disables metrics for that path (no `in_total`, no `ok/fail`, invisible in Grafana). Test callers can pass `RpcRequestMetrics{}` explicitly — please drop the default from all three signatures.

Bigger picture: the service impl shouldn't need to know about admission counters at all. Consider having APIService own the `shared_ptr<RpcRequestMetrics>` end-to-end and pass a `std::function<void(bool ok)> on_complete` into the service. That removes `RpcRequestMetrics` from three header signatures, kills `RpcRequestMetrics::detach()` (which exists only to paper over dual ownership), removes the duplicated 10-line callback lambdas in the two impls, and makes `Call::mark_rpc_failed()` unneeded.

xllm/api_service/api_service.cpp
```diff
@@ -200,6 +201,7 @@ void APIService::CompletionsHttp(::google::protobuf::RpcController* controller,
       google::protobuf::Arena::CreateMessage<proto::CompletionResponse>(arena);
 
   auto ctrl = reinterpret_cast<brpc::Controller*>(controller);
```
ustcfy, line 203: `custom-code-style.md` §3 requires `static_cast` — `google::protobuf::RpcController*` → `brpc::Controller*` is a well-defined downcast. You already converted a couple of sites in this PR to `brpc::Controller* ctrl = static_cast<brpc::Controller*>(controller);`. Please convert the remaining ~12 `reinterpret_cast<brpc::Controller*>` sites in this file (Completions non-http, CompletionsHttp, Sample, SampleHttp, ChatCompletions, ChatCompletionsHttp, Embeddings, embedding, ImageGenerationHttp, VideoGenerationHttp, RerankHttp, handle_anthropic_messages, AnthropicMessagesHttp) so the file is consistent.

xllm/api_service/api_service.cpp
```diff
@@ -152,42 +155,40 @@ void APIService::Completions(::google::protobuf::RpcController* controller,
...
         const_cast<proto::CompletionRequest*>(request),
         response,
-        arena != nullptr);
+        arena != nullptr,
+        /*is_http_request=*/false,
+        std::move(rpc_metrics));
     rec_completion_service_impl_->process_async(call);
+  } else {
+    ctrl->SetFailed("Completion service is not available on this server");
+    LOG(ERROR) << "Completion service is not available on this server";
```
ustcfy, line 183: This `SetFailed + LOG(ERROR)` pair with a duplicated string literal is repeated 4× in this PR (Completions non-http, CompletionsHttp, ChatCompletions, ChatCompletionsHttp). Consider a file-local anonymous-namespace helper:
```cpp
void fail_with_log(brpc::Controller* ctrl, const char* msg) {
  ctrl->SetFailed(msg);
  LOG(ERROR) << msg;
}
```
Keeps the log message and the client-visible failure string in one place.

xllm/api_service/api_service.cpp
```diff
@@ -935,6 +916,7 @@ void handle_anthropic_messages(std::unique_ptr<AnthropicServiceImpl>& service,
                                brpc::Controller* ctrl,
                                const proto::HttpRequest* request,
                                proto::HttpResponse* response) {
+  RpcRequestMetrics rpc_metrics(ctrl);
```
ustcfy, line 919: Style nit on the enclosing `handle_anthropic_messages` signature (line 914): passing `ClosureGuard& done_guard` so the callee can call `.release()` is an implicit ownership handoff — the return-ownership contract is invisible at the call site. Consider having the callee return `google::protobuf::Closure*` (or `std::optional<...>`) and letting the caller decide what to do with it. Explicit ownership is easier to audit.

xllm/api_service/chat_service_impl.cpp
```diff
@@ -612,13 +614,25 @@ void ChatServiceImpl::process_rec_chat_request(std::shared_ptr<ChatCall> call) {
...
-    const proto::ChatRequest* request) {
+void ChatServiceImpl::process_async_rpc_impl(const proto::ChatRequest* request,
+                                             RpcRequestMetrics rpc_metrics) {
+  rpc_metrics.detach();
+  std::shared_ptr<RpcRequestMetrics> metrics =
+      std::make_shared<RpcRequestMetrics>(std::move(rpc_metrics));
   const auto& service_request_id = request->service_request_id();
   const auto& target_xservice_addr = request->source_xservice_addr();
-  auto callback = [master = master_](const RequestOutput& req_output) -> bool {
+  auto callback = [master = master_,
```
ustcfy, line 624: This ~10-line callback body is duplicated verbatim in `completion_service_impl.cpp` (only the surrounding types differ). Please extract to a shared helper:
```cpp
template <class Master>
auto make_rpc_response_callback(
    Master* master, std::shared_ptr<RpcRequestMetrics> metrics);
```
One place to update "what counts as a failure."

Efficiency note: `std::function` copies the lambda; with two captures (raw `master` + `shared_ptr<RpcRequestMetrics>`), the capture likely exceeds SBO and heap-boxes the lambda → **potentially 2 allocs per request**. Bundling `master + metrics` into a single `shared_ptr<struct RpcCallbackState>` guarantees 1 alloc with better cache locality. (If APIService takes over metric ownership per the `api_service_impl.h` comment, this concern disappears entirely.)

Corner case distinct from the earlier codex "handle_rpc_response returned false" comment (which this PR now handles): if `master_->handle_request` (line ~708 below) accepts the request but the master silently drops it and never invokes `callback` at all, the `metrics` shared_ptr eventually dies via the master's callback container and the dtor records **ok** — masking a real failure. Either document "master promises exactly-once callback" on `handle_request`, or `metrics->mark_failed()` before `handle_request` returns so a later "ok" record requires an explicit success. Same concern in `completion_service_impl.cpp`.

xllm/api_service/stream_call.h
```diff
@@ -83,6 +93,8 @@ class StreamCall : public Call {
   }
 
   bool write_and_finish(Response& response) {
+    CHECK(controller_ != nullptr)
```
ustcfy, line 96: This CHECK (and the siblings at ~115 and in `AnthropicCall::write_and_finish`) encodes at runtime what the class already tracks via `stream_`. If any caller ever wires a stream request through a non-stream API surface, the whole server `LOG(FATAL)`s.

Consider splitting `StreamCall` into sibling types `NonStreamCall` (owns controller through dtor) and `StreamingCall` (releases controller at handshake, owns `pa_`), both inheriting from a thinner `Call`. The distinction becomes compile-time and every one of these CHECKs disappears. Not blocking this PR, but worth a follow-up — the current one-class-two-modes shape is the *cause* of the fragile CHECKs the PR added.

tests/api_service/rpc_request_metrics_test.cpp
```diff
@@ -0,0 +1,190 @@
...
+  {
+    RpcRequestMetrics first(&controller);
+    RpcRequestMetrics second(std::move(first));
+  }
+
+  EXPECT_EQ(COUNTER_VALUE(server_request_in_total), in_before + 1.0);
+  EXPECT_EQ(COUNTER_VALUE(server_request_total_ok), ok_before + 1.0);
+}
+
+TEST(RpcRequestMetricsTest, FinishThenDestroyControllerDoesNotUAF) {
```
ustcfy, line 110: Nit: without ASAN this test only verifies that `finish()` clears `controller_` — not the actual UAF guarantee. Either rename to `FinishClearsControllerReference`, or add a comment noting ASAN is required for real UAF coverage.

PR #2339, opened 2026-09-15
feat: support GLM-5.3 Python MTP execution.

edge-nullptr:
## Summary

- Add the Python GLM-5.3 MTP model and executor path.
- Preserve GLM DSA layer metadata and handle mismatched target metadata safely.
- Allow Python GLM MTP speculative startup when the MTP draft path is available.
- Add registry, model-loader, executor, and MTP coverage.

## Validation

- pre-commit checks passed for the changed C++ and Python files.
- Runtime validation should be run after building the NPU target.


xllm/python/models/glm5_2.py
```diff
@@ -739,21 +750,38 @@ def select_qli(
...
-            # only the padded K rows obey the equal-size CP gather contract.
+            # Only the padded K rows obey the equal-size CP gather contract.
             k = cp_gather_kv(k, ctx.cp_context).contiguous()
 
         index_cache = ctx.index_cache
         index_cache_scale = ctx.index_cache_scale
-        use_quant_indexer = index_cache.dtype == torch.int8 and index_cache_scale is not None
         k_scale = None
-        if use_quant_indexer:
+        if index_cache.dtype == torch.int8 and index_cache_scale is not None:
```
XuZhang99, line 759: why remove `use_quant_indexer`?

edge-nullptr: `use_quant_indexer` was only used once inside `_update_index_cache()`, so that helper now inlines the same predicate. The variable is still retained in `select_qli()`, where it selects the quantized versus floating-point indexer kernel. This is a readability-only refactor; there is no intended behavior change.

edge-nullptr: Restored the local `use_quant_indexer` name in `_update_index_cache()` as well for consistency with `select_qli()`. This is a style-only follow-up in commit `8e2b8f51`; behavior is unchanged.

XuZhang99 (dismissed):

XuZhang99 (approved):

